PR #13, opened 2021-10-08
get metamask to stop retrying, by hijacking eth_getTransactionCount

metachris:


bertmiller (approved):
Tested locally and successfully:
- dropped an underpriced transaction
- relayed and landed a well priced transaction

Great job coding this up on short notice

PR #27, opened 2021-10-20
MetaMask fix 2

metachris:
Fix 2 to help unstuck transactions:

* Also includes some necessary refactoring, including remembering transactions

Steps:

* on `sendRawTransaction`: memorize time the tx was received
* on `eth_getTransactionReceipt`: if result is `null` and tx submission time is >16 min:
  * call `eth_getBundleStatusByTransactionHash` on BE
  * if `result.status` is `FAILED_BUNDLE` then blacklist the EOA so subsequent calls to `eth_getTransactionCount` return a high nonce

Note: 16 minutes is chosen as the time interval because MetaMask increases the time between retries every block with a limit of 15 minutes. If MetaMask has not sent a retry in 15 minutes and the BE has a status of failed for the tx we can confidently conclude that the tx has been dropped.

### DIFF
--- a/server/request.go
+++ b/server/request.go
@@ -152,8 +152,8 @@ func (r *RpcRequest) process() {
 		readJsonRpcSuccess, proxyHttpStatus, jsonResp := r.proxyRequestRead(r.defaultProxyUrl)
 
 		// After proxy, perhaps check backend [MM fix #3 step 2]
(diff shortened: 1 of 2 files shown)


bertmiller:
LGTM! Testing with two transactions on Mainnet with staging.

PR #31, opened 2021-11-18
Major cleanup, Improved Metamask failed tx handling

metachris:
* Improved Metamask failed tx handling
  * Added support for `DEBUG_DONT_SEND_RAWTX` environment variable, which will fake sending rawTransactions (returns the tx hash like on success, but doesn't forward it to relay/mempool). This triggers the Metamask failed transaction flow immediately, without needing to wait for a tx to expire.
  * Block transactions with 1e9+ nonce with an error that leads Metamask to drop a transaction immediately
  * Only triggering nonce fix on the latest transaction of a specific sender address
* Introduction of a state struct which ill be used to put state into Redis in the future
* Code cleanup: remove artifacts of legacy infra, dead codepaths, some simplifications
* Allow immediate resubmission of a tx to the relay once state is "FAILED" (don't need to wait for any additional amount of time)
* Better handle errors from the relay


metachris:
deployed on staging

bertmiller:
Pulled data and successfully landed a tx. You said that the dropping conditions work, right? I have a low base fee tx out right now and will check it after lunch.

server/state.go
```diff
@@ -0,0 +1,75 @@
...
+		txHash:    strings.ToLower(txHash),
+	}
+}
+
+// todo: put into redis
+type GlobalState struct {
+	txForwardedToRelay  map[string]time.Time      // key: txHash
+	accountWithNonceFix map[string]*nonceFix      // key: txFrom
+	userLatestTx        map[string]StringWithTime // key: txFrom, value: txHash
+	txToUser            map[string]StringWithTime // key: txHash, value: txFrom
```
bertmiller, line 26: nitpicking but `txHashToUser` and `userLatestTxHash` are more descriptive. Otherwise someone might expect the transaction object

server/request.go
```diff
@@ -136,7 +132,7 @@ func (r *RpcRequest) process() {
 	}
 
 	r.log("JSON-RPC method: %s ip: %s", r.jsonReq.Method, r.ip)
-	MetaMaskFix.CleanupStaleEntries()
+	State.cleanup()
```
bertmiller, line 135: Do we need to do this on every rpc request? Does it matter? Alternative is to only call on eth_sendRawTransaction

bertmiller (approved):
LGTM

server/request.go
metachris: well spotted, thanks!

server/state.go
metachris: Good point :)

metachris:
> There might be some weird conditions around if users are using us from two wallets, but I think that's mostly fine

Hmm I think this should be fine, since "user" references a wallet, and all the metamask fixes are per EOA. Am I missing something?

> Having state be pulled out and separate makes sense to me, will allow us to run multiple things in parallel in the future I think

Exactly! We'll put the state into Redis eventually.

PR #33, opened 2021-11-19
failed getTxReceipt after user received 4x wrong nonce

metachris:
This PR implements an error response on `getTransactionReceipt` after Metamask has received the wrong nonce 4x (at which point Metamask shows the tx as dropped).

An error response on `getTransactionReceipt` allows websites/dapps like matcha.xyz to show that the transaction has actually failed, instead of being stuck in pending, because a proxied getTransactionReceipt still returns null (the tx never made it to mempool).

As an alternative, we could return an error on `getTransactionReceipt` immediately once the transaction API returns failed state, but in this case Metamask would still show the tx as pending for a minute or two, until it has received the nonces 4x.


metachris:
deployed on staging

metachris:
Will be superseded shortly

PR #38, opened 2021-11-22
Putting state into Redis

metachris:
and some refactorings/bugfixes


server/request_intercepts.go
```diff
@@ -19,48 +21,64 @@ func (r *RpcRequest) check_post_getTransactionReceipt(jsonResp *JsonRpcResponse)
...
-
-	// Abort if transaction wasn't submitted before
-	txFrom, txFound := State.txHashToUser[txHashLower]
-	if !txFound {
-		return
-	}
-	txFromLower := txFrom.s
-
-	// Abort if not the latest transaction by user
-	latestTxHash, latestFound := State.userLatestTxHash[txFromLower]
```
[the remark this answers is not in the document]
metachris: Yes, this was in place for special logic on only the latest transaction of a user (remove the nonce-fix if the last tx succeeded). But that logic was faulty in first place and wasn't triggered, and is removed now until perhaps another time and better implementation.

server/request.go
```diff
@@ -392,44 +400,154 @@ func (r *RpcRequest) _writeRpcResponse(res *JsonRpcResponse) {
...
+
+	if DebugDontSendTx {
+		r.log("faked sending tx to relay, did nothing")
+		r.writeRpcResult(initialTxHash)
 		return
 	}
 
-	if backendResp.Error != nil {
-		r.logError("[sendTxToRelay] relay returned an error for %s: %s", txHash, backendResp.Error.Message)
+	cancelPrivTxArgs := flashbotsrpc.FlashbotsCancelPrivateTransactionRequest{TxHash: initialTxHash}
```
bertmiller, line 538: Just a reminder to merge the private tx updates on your flashbots repo

server/request.go
```diff
@@ -208,41 +212,45 @@ func (r *RpcRequest) handle_sendRawTransaction() {
...
-			// original tx was sent to relay
-			r.log("[cancel-tx] sending to relay for %s", txFromLower)
-			needsProtection = true
-		} else {
-			// original tx was sent to mempool, or not seen in rpc-endpoint
-			r.log("[cancel-tx] sending to mempool for %s", txFromLower)
-			needsProtection = false
+	// Check for cancellation-tx
+	if len(r.tx.Data()) <= 2 && txFromLower == strings.ToLower(r.tx.To().Hex()) {
+		requestDone := r.handleCancelTx() // returns true if tx was cancelled at the relay and response has been sent to the user
```
bertmiller, line 246: What's the behaviour on the user's side here? Their cancellation tx looks successfully sent, right?

We could immediately fail that tx, although that might be confusing. I guess this returning a success and dropping them is best.

bertmiller (commented):
Couple comments, hadn't finished yet but looking good so far

server/request.go
```diff
@@ -392,44 +400,154 @@ func (r *RpcRequest) _writeRpcResponse(res *JsonRpcResponse) {
...
+		r.writeRpcResult(cancelTxHash)
+		return true
+	}
+
+	r.log("[cancel-tx] sending to relay: %s for %s/%d", initialTxHash, txFromLower, r.tx.Nonce())
+
+	if DebugDontSendTx {
+		r.log("faked sending tx to relay, did nothing")
+		r.writeRpcResult(initialTxHash)
 		return
```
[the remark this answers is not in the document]
metachris: What do you mean with "explicit return"? This function doesn't return any usable value, so `return` just exits 🤔

server/request.go
metachris: The cancel-tx should succeed, and the initial tx (the one that was cancelled) should receive failed on next getTransactionReceipt

server/request.go
```diff
@@ -392,44 +400,154 @@ func (r *RpcRequest) _writeRpcResponse(res *JsonRpcResponse) {
...
+		}
+		return
+	}
+
+	r.writeRpcResult(txHash)
+	r.log("[sendTxToRelay] sent %s", txHash)
+}
+
+// Sends cancel-tx to relay as cancelPrivateTransaction, if initial tx was sent there too.
+func (r *RpcRequest) handleCancelTx() (requestCompleted bool) {
```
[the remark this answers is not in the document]
metachris: Good point. I've tried doing this just now (splitting into check and send function), but the issue I've encountered is that both need to call Redis `GetTxHashForSenderAndNonce`, to get the tx hash for sender+nonce, which doubles all the error handling around Redis, which I'd prefer to avoid in this case.

server/request.go
metachris: 👍 

https://github.com/metachris/flashbotsrpc/pull/4

bertmiller:
This was a different PR, I think we should change the name of `shouldSendTxToRelay`. It's a little confusing what that does because it's triggered in a function called "sendTxToRelay"

How about checkTxStatusAtRelay?

server/request_intercepts.go
```diff
@@ -71,18 +89,29 @@ func (r *RpcRequest) intercept_mm_eth_getTransactionCount() (requestFinished boo
...
+	if !nonceFixInPlace {
 		return false
 	}
 
 	// Intercept max 4 times (after which Metamask marks it as dropped)
-	nonceFix.numTries += 1
-	if nonceFix.numTries > 4 {
+	numTimesSent += 1
+	if numTimesSent > 4 {
 		return
```
bertmiller, line 105: Should this return a bool?

bertmiller:
Altogether LGTM, I didn't see anything that should be a showstopper!

Would be good for us to take a step back add some issues for others to contribute to, like we discussed

server/request_intercepts.go
metachris: done

metachris:
> This was a different PR, I think we should change the name of `shouldSendTxToRelay`. It's a little confusing what that does because it's triggered in a function called "sendTxToRelay"
> 
> How about checkTxStatusAtRelay?

But this function does more than checking at the relay -- it first checks the local cache if this tx has been submitted before, and if yes only then checks status at relay. For this reason I think `checkTxStatusAtRelay` is less descriptive than `shouldSendTxToRelay`. Thinking about a better name, open to changing it :)

server/request.go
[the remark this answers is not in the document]
Ruteri: I may have misread, or it's no longer relevant - either way consider resolved.

metachris:
Deployed to staging

server/request.go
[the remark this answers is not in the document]
Ruteri: Sure, if it's not easily splitable and you tried then I don't have any more inputs

PR #44, opened 2021-12-08
relay checks: correct nonce, large tx to whitelisted addresses

metachris:


server/request.go
```diff
@@ -244,7 +245,46 @@ func (r *RpcRequest) sendTxToRelay() {
...
+	err = json.Unmarshal(_res.Result, &_nonceStr)
+	if err != nil {
+		r.logError("[sendTxToRelay] eth_getTransactionCount unmarshall failed: %v - result: %s", err, _res.Result)
+		r.writeHeaderStatus(http.StatusInternalServerError)
+		return
+	}
+	_nonceStr = strings.Replace(_nonceStr, "0x", "", 1)
+	_nonceBigInt := new(big.Int)
+	_nonceBigInt.SetString(_nonceStr, 16)
+	if r.tx.Nonce() != _nonceBigInt.Uint64() {
```
jparyani, line 272: This will fail txs when you have a pending tx waiting. Should we be storing nonces of pending outgoing txs and use the max of that/eth_getTransactionCount.

metachris: good catch! yes we should :)

metachris: Updated the code, pls review again :)

server/redisstate.go
```diff
@@ -29,6 +29,10 @@ var RedisExpiryNonceFixForAccount = time.Duration(24 * time.Hour)
 var RedisPrefixSenderOfTxHash = RedisPrefix + "txsender-of-txhash:"
 var RedisExpirySenderOfTxHash = time.Duration(24 * time.Hour) // 1 day
 
+// Remember nonce of pending user tx
+var RedisPrefixSenderMaxNonce = RedisPrefix + "txsender-pending-max-nonce:"
+var RedisExpirySenderMaxNonce = time.Duration(24 * time.Hour) // 1 day
```
jparyani, line 34: 1 day might be too long? Shouldn't these expire within 25 blocks so roughly 10-20 minutes?

jparyani (approved):

server/whitelist.go
```diff
@@ -21,3 +21,7 @@ func isOnFunctionWhiteList(data string) bool {
 		return false
 	}
 }
+
+var allowedLargeTxTargets = map[string]bool{
+	"0x737901bea3eeb88459df9ef1be8ff3ae1b42a2ba": true, // Aztech rollup contract
```
obadiaa, line 26:
```suggestion
	"0x737901bea3eeb88459df9ef1be8ff3ae1b42a2ba": true, // Aztec rollup contract
```

server/redisstate.go
metachris: Perhaps one or two hours would be best, because it needs to take into account nonces from tx in the public mempool too

server/whitelist.go
metachris: thx 👍

Ruteri (approved):

PR #45, opened 2021-12-10
parse bundle-id from URL params & attempt to store in redis

zeroXbrock:
DON'T MERGE. PR just for code comparison & concept exploration.

### DIFF
--- a/server/redisstate.go
+++ b/server/redisstate.go
@@ -190,7 +190,9 @@ func (s *RedisState) AddTxToBundle(bundleId string, signedTx string) error {
 	// append new tx if array exists, otherwise just create a new array
 	// txs := make([]string, 1)
 	// txs[0] = signedTx
-	err := s.RedisClient.Set(context.Background(), key, signedTx, RedisExpiryBundleTransactions).Err()
+	// TODO: how to store arrays in redis w/ go? idk any ser/de methods in Go
+	txs := signedTx
+	err := s.RedisClient.Set(context.Background(), key, txs, RedisExpiryBundleTransactions).Err()
 	return err
 }
 
--- a/server/redisstate.go
+++ b/server/redisstate.go
@@ -29,6 +29,10 @@ var RedisExpiryNonceFixForAccount = time.Duration(24 * time.Hour)
 var RedisPrefixSenderOfTxHash = RedisPrefix + "txsender-of-txhash:"
 var RedisExpirySenderOfTxHash = time.Duration(24 * time.Hour) // 1 day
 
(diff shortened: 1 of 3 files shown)


zeroXbrock:
NOT FINISHED. PR just for code comparison & conceptual exploration.

server/request_sendrawtx.go
```diff
@@ -69,6 +69,9 @@ func (r *RpcRequest) handle_sendRawTransaction() {
 	// Check if transaction needs protection
 	needsProtection := r.doesTxNeedFrontrunningProtection(r.tx)
 
+	// Check if tx is being cached for later bundling
+	cachingTx := r.isCachingBundleTx()
```
metachris, line 73:
```suggestion
	_, cachingTx := r.req.URL.Query()["bundle-id"]
```

How about we remove the whole `isCachingBundleTx` function?

server/redisstate.go
```diff
@@ -172,6 +180,22 @@ func (s *RedisState) GetSenderOfTxHash(txHash string) (txSender string, found bo
...
+// Enable lookup of tx bundles by bundle ID
+//
+func (s *RedisState) AddTxToBundle(bundleId string, signedTx string) error {
+
+	key := RedisKeyBundleTransactions(bundleId)
+	// TODO: get existing array if it exists
+	// append new tx if array exists, otherwise just create a new array
+	// txs := make([]string, 1)
+	// txs[0] = signedTx
+	// TODO: how to store arrays in redis w/ go? idk any ser/de methods in Go
```
metachris, line 193: https://redis.io/commands/lpush

metachris:
I feel that using the term `bundle-id` is misleading. "bundle" and "bundle-id" is already used and refers specifically to Flashbots Relay bundles and the id of them. I suggest we change the term to `cacheId`, and the query arg to `cid`, like `rpc.flashbots.net?cid=foo`

metachris:
I've pushed a PR with updates to this PR: https://github.com/zeroXbrock/rpc-endpoint/pull/1

bertmiller:
"Whitehat-ID" could be an alternative

metachris:
> "Whitehat-ID" could be an alternative

The issue is that mistypings will fail silently, and the longer the query arg is the easier to mistype. Users will think they are all set and wonder why it's not working. Perhaps best would simply be a 1-letter arg, like `?x=foo`?

metachris:
Created the branch `whitehat-txcache` with this PR merged in, and others opened to it. Let's work against that branch until we merge.

Summary PR in #48

PR #49, opened 2021-12-12
Whitehat txcache

metachris:
Todo:

- [x] Rename `bundle-id` to something better
- [ ] Test & merge #50 
- [x] Add maximum number of 15 tx per key (using [ltrim](https://redis.io/commands/ltrim))
- [x] Intercept [`eth_getBalance`](https://infura.io/docs/ethereum/json-rpc/eth-getBalance) and return 1 ETH
- [x] Add a method for users to retrieve the cached keys (simple REST API?)
- [x] Add e2e tests
- [x] Test on staging

### DIFF
--- a/server/redisstate.go
+++ b/server/redisstate.go
@@ -33,6 +33,10 @@ var RedisExpirySenderOfTxHash = time.Duration(24 * time.Hour) // 1 day
 var RedisPrefixSenderMaxNonce = RedisPrefix + "txsender-pending-max-nonce:"
 var RedisExpirySenderMaxNonce = time.Duration(2 * time.Hour)
 
+// Enable lookup of bundle txs by bundleId
+var RedisPrefixWhitehatBundleTransactions = RedisPrefix + "tx-for-whitehat-bundle:"
+var RedisExpiryWhitehatBundleTransactions = time.Duration(24 * time.Hour) // 1 day
+
 // // Enable lookup of last privateTransaction-txHash sent by txFrom
 // var RedisPrefixLastPrivTxHashOfAccount = RedisPrefix + "last-txhash-of-txsender:"
 // var RedisExpiryLastPrivTxHashOfAccount = time.Duration(24 * time.Hour) // 1 day
@@ -55,6 +59,10 @@ func RedisKeySenderOfTxHash(txHash string) string {
 
 func RedisKeySenderMaxNonce(txFrom string) string {
 	return RedisPrefixSenderMaxNonce + strings.ToLower(txFrom)
+}
+
+func RedisKeyWhitehatBundleTransactions(bundleId string) string {
+	return RedisPrefixWhitehatBundleTransactions + strings.ToLower(bundleId)
 }
 
 // func RedisKeyLastPrivTxHashOfAccount(txFrom string) string {
@@ -178,6 +186,39 @@ func (s *RedisState) GetSenderOfTxHash(txHash string) (txSender string, found bo
 	}
 
 	return strings.ToLower(txSender), true, nil
+}
+
+//
+// Enable lookup of tx bundles by bundle ID
+//
+func (s *RedisState) AddTxToWhitehatBundle(bundleId string, signedTx string) error {
+	key := RedisKeyWhitehatBundleTransactions(bundleId)
+
+	// Add item
+	err := s.RedisClient.LPush(context.Background(), key, signedTx).Err()
+	if err != nil {
+		return err
+	}
+
+	// Set expiry
+	err = s.RedisClient.Expire(context.Background(), key, RedisExpiryWhitehatBundleTransactions).Err()
+	if err != nil {
+		return err
+	}
+
+	// Limit to 15 entries
+	err = s.RedisClient.LTrim(context.Background(), key, 0, 15).Err()
+	return err
+}
+
+func (s *RedisState) GetWhitehatBundleTx(bundleId string) ([]string, error) {
+	key := RedisKeyWhitehatBundleTransactions(bundleId)
+	return s.RedisClient.LRange(context.Background(), key, 0, -1).Result()
+}
+
+func (s *RedisState) DelWhitehatBundleTx(bundleId string) error {
+	key := RedisKeyWhitehatBundleTransactions(bundleId)
+	return s.RedisClient.Del(context.Background(), key).Err()
 }
 
 //
--- a/server/request_sendrawtx.go
+++ b/server/request_sendrawtx.go
@@ -75,6 +75,19 @@ func (r *RpcRequest) handle_sendRawTransaction() {
 	// Check if transaction needs protection
 	needsProtection := r.doesTxNeedFrontrunningProtection(r.tx)
 
+	// If users specify a bundle ID, cache this transaction
+	if r.isWhitehatBundleCollection {
+		r.logger.log("[WhitehatBundleCollection] adding tx to bundle %s txData: %s", r.whitehatBundleId, r.rawTxHex)
+		err = RState.AddTxToWhitehatBundle(r.whitehatBundleId, r.rawTxHex)
+		if err != nil {
+			r.logger.logError("[WhitehatBundleCollection] AddTxToWhitehatBundle failed:", err)
+			r.writeRpcError("[WhitehatBundleCollection] AddTxToWhitehatBundle failed:", types.JsonRpcInternalError)
+			return
+		}
+		r.writeRpcResult(r.tx.Hash().Hex())
+		return
+	}
+
 	// Check for cancellation-tx
 	if len(r.tx.Data()) <= 2 && txFromLower == strings.ToLower(r.tx.To().Hex()) {
 		requestDone := r.handleCancelTx() // returns true if tx was cancelled at the relay and response has been sent to the user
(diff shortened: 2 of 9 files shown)


server/request_sendrawtx.go
```diff
@@ -69,6 +69,13 @@ func (r *RpcRequest) handle_sendRawTransaction() {
 	// Check if transaction needs protection
 	needsProtection := r.doesTxNeedFrontrunningProtection(r.tx)
 
+	// If users specify a bundle ID, cache this transaction
+	if cacheId := r.req.URL.Query().Get("bundle-id"); cacheId != "" {
```
metachris, line 73: We should rename `bundle-id` to something better - let's discuss this in Discord

metachris:
ping @zeroXbrock

server/request_sendrawtx.go
metachris: After some more thinking, how about just `?bundle=123`?

zeroXbrock (commented):
Awesome -- it works great. Just had a couple comments on the false ETH balance value, and a quick question about error handling.

Last thing we need to address before this is integrated is how to signal that a bundle is done being assembled, and how to send the bundle. The signal could simply be a call to a specific smart contract address. After that, how do we pass the cached bundle to the relay?

server/request.go
```diff
@@ -130,6 +136,9 @@ func (r *RpcRequest) process() {
 		} else if r.jsonReq.Method == "net_version" { // don't need to proxy to node, it's always 1 (mainnet)
 			r.writeRpcResult("1")
 			return
+		} else if r.isWhitehatBundleCollection && r.jsonReq.Method == "eth_getBalance" {
+			r.writeRpcResult("0xde0b6b3a7640000") // 1 eth in gwei
```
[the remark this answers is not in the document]
metachris: > Also, new idea, what if we only return a false balance when there are transactions in redis for this bundle ID?

The problem would be that users perhaps can't do the first transaction.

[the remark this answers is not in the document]
metachris: Updated the code to return 1337 ETH + (0.01 ETH * number of cached tx)

server/request_sendrawtx.go
```diff
@@ -69,6 +69,17 @@ func (r *RpcRequest) handle_sendRawTransaction() {
 	// Check if transaction needs protection
 	needsProtection := r.doesTxNeedFrontrunningProtection(r.tx)
 
+	// If users specify a bundle ID, cache this transaction
+	if r.isWhitehatBundleCollection {
+		r.log("[WhitehatBundleCollection] adding tx to bundle %s txData: %s", r.whitehatBundleId, r.rawTxHex)
+		err = RState.AddTxToWhitehatBundle(r.whitehatBundleId, r.rawTxHex)
+		if err != nil {
+			r.logError("[WhitehatBundleCollection] AddTxToWhitehatBundle failed:", err)
```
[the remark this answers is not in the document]
metachris: Good catch, done

metachris:
> Last thing we need to address before this is integrated is how to signal that a bundle is done being assembled, and how to send the bundle. The signal could simply be a call to a specific smart contract address. After that, how do we pass the cached bundle to the relay?

Wouldn't this be part of your backend/frontend? Like have a button to take all the tx and create a bundle and send it to the relay? Not sure this should be part of the RPC endpoint, probably it shouldn't.

server/request.go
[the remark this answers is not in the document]
zeroXbrock: > > Also, new idea, what if we only return a false balance when there are transactions in redis for this bundle ID?
> 
> The problem would be that users perhaps can't do the first transaction.

ah, good point

zeroXbrock:
> > Last thing we need to address before this is integrated is how to signal that a bundle is done being assembled, and how to send the bundle. The signal could simply be a call to a specific smart contract address. After that, how do we pass the cached bundle to the relay?
> 
> Wouldn't this be part of your backend/frontend? Like have a button to take all the tx and create a bundle and send it to the relay? Not sure this should be part of the RPC endpoint, probably it shouldn't.

Replied on discord, but for posterity, yes. Much better design than my previous proposition.

zeroXbrock:
LGTM

odyslam:
I take that the feature is not yet active? Is there any plans/timeline for Goerli support? As I test a nice smol tool for this feature and testing on mainnet is somewhat expensive.

metachris:
> I take that the feature is not yet active? Is there any plans/timeline for Goerli support? As I test a nice smol tool for this feature and testing on mainnet is somewhat expensive.
> 
> btw, thank you so much for this. Such a great enable for easier whitehats

This code is deployed on staging (`rpc-staging.flashbots.net`). We don't have Protect on Goerli at the moment.

odyslam:
Thanks @metachris ! I have been testing and playing with it and I think it doesn't reset with every new bundleId I generate.

odyslam:
Hey good friends, 

I have been playing around with the feature and I have strange behavior, where I submit a single transaction and the bundle has like 16 when I `get` it. 

bundle-id: `df878646-0625-4164-ad99-1389fafbf27f`

I have verified that the array `transactions` has the correct number of objects before I start iterating and send the transactions: https://github.com/odyslam/ethtools/blob/dbfebee820e303bd027d0dd005693df138119f8f/index.js#L154

metachris:
@zeroXbrock have you been seeing something similar, or is everything working as expected for you?

odyslam:
btw, here is how I surface the functionality if you want to give it a spin: http://ethtools.odyslam.com/flashbots

Interestingly, all the txs that I sent to protect, stay in metamask's queue, so when I switch metamask back to the default mainnet, it will batch send all the txs that were sent to protect to be batched into bundles. T

zeroXbrock:
> Hey good friends,
> 
> I have been playing around with the feature and I have strange behavior, where I submit a single transaction and the bundle has like 16 when I `get` it.
> 
> bundle-id: `df878646-0625-4164-ad99-1389fafbf27f`
> 
> I have verified that the array `transactions` has the correct number of objects before I start iterating and send the transactions: https://github.com/odyslam/ethtools/blob/dbfebee820e303bd027d0dd005693df138119f8f/index.js#L154

@odyslam I'm not sure why you're seeing so many transactions. Are you using a different ID for each bundle?

> @zeroXbrock have you been seeing something similar, or is everything working as expected for you?

I just ran a quick sanity check on localhost and it seems to be working fine. Steps to reproduce:

**Spin up RPC**

```sh
# in rpc-endpoint/
make
./rpc-endpoint --signingKey <random private key here>
```

**Set metamask URL** to `http://localhost:9000?bundle=13` (not using UUID just for readability).

**Send any tx in Metamask.** I just sent some ETH to another account; all I want to do is add one tx to bundle `13`.

**Check transactions**

```sh
curl http://localhost:9000/bundle?id=13 | jq
```

I see one transaction in my instance:
```json
{
  "bundleId": "13",
  "rawTxs": [
    "0x02f876827..."
  ]
}
```

I then added another network to metamask with a new bundle ID `http://localhost:9000?bundle=42`, ran through the same process and only saw one transaction, which is what we expect. Also verified that both bundles stayed intact. 

@odyslam Could you provide more information about your experience? How and when did the extra transactions show up? Did you just send one tx, and then find 16 upon GETting the bundle, or how did that go down? Were any of the transactions mined?

odyslam:
Hey @zeroXbrock 

Thanks so much for this thorough reply. I was fortunate and debugged this over discord with @epheph . The final verdict is that probably metamask is sending transactions that it shouldn't. Older transactions that is. 

The solution for me was to verify the received bundle and remove any txs that is not verified with the set of unsigned transactions before I send them to the cache.

Quite interestingly, the number 16 will not increment as I send more txs through metamask and it will not stop surfacing even if I reset the account.

To remove the not valid ones I check for equality in the fields (data, to, value, estimatedbasefee). The last one is the best cause it's a very specific number that seems to change all the time.

server/redisstate.go
```diff
@@ -180,6 +188,39 @@ func (s *RedisState) GetSenderOfTxHash(txHash string) (txSender string, found bo
 	return strings.ToLower(txSender), true, nil
 }
 
+//
+// Enable lookup of tx bundles by bundle ID
+//
+func (s *RedisState) AddTxToWhitehatBundle(bundleId string, signedTx string) error {
+	key := RedisKeyWhitehatBundleTransactions(bundleId)
+
```
bhakiyakalimuthu, line 196: small suggestion. declare variable as ctx := context.Background(); then It can be reused wherever required

metachris:
> @metachris do we delete transactions from redis once the bundle is mined?

No, the rpc endpoint has no notion of mined bundles. The content can be deleted with the DELETE API call, or it expires after 24 hours.

server/redisstate.go
metachris: Not exactly sure what you mean - could you perhaps give an example?

bhakiyakalimuthu: apologies for not being clear. simple change
`ctx := context.Background()
err := s.RedisClient.LPush(ctx, key, signedTx).Err()
	if err != nil {
		return err
	}

	// Set expiry
	err = s.RedisClient.Expire(ctx, key, RedisExpiryWhitehatBundleTransactions).Err()
	if err != nil {
		return err
	}

	// Limit to 15 entries
	err = s.RedisClient.LTrim(ctx, key, 0, 15).Err()
	return err
}`

PR #50, opened 2021-12-12
whitehat-txcache: max number of redis keys for cached tx

metachris:


server/redisstate.go
```diff
@@ -189,23 +181,10 @@ func (s *RedisState) AddTxToBundle(bundleId string, signedTx string) error {
...
-// 	txHash, err = s.RedisClient.Get(context.Background(), key).Result()
-// 	if err == redis.Nil { // not found
-// 		return "", false, nil
-// 	} else if err != nil {
-// 		return "", false, err
-// 	}
-
-// 	return strings.ToLower(txHash), true, nil
-// }
+func (s *RedisState) GetNumberOfBuncleCacheKeys() (uint64, error) {
```
Ruteri, line 184: Typo, should probably `GetNumberOfBundleCacheKeys`

server/request_sendrawtx.go
```diff
@@ -71,6 +71,12 @@ func (r *RpcRequest) handle_sendRawTransaction() {
 
 	// If users specify a bundle ID, cache this transaction
 	if cacheId := r.req.URL.Query().Get("bundle-id"); cacheId != "" {
+		if !r.allowTxCache {
```
Ruteri, line 74: I wonder if this check should cause rpc calls to fail, since users can't even check if it's enabled

metachris: yes it will make it fail. users wouldn't check because they are using metamask anyway to create bundles. don't think that failing really is an issue 🤔

metachris:
reopening later

PR #52, opened 2021-12-22
skip serializing result if not set

sunce86:
Without `omitempty`, `result` field is serialized even when not populated, causing jsonrpc_core crate to fail deserializing failure response:

https://docs.rs/jsonrpc-core/latest/jsonrpc_core/types/response/struct.Failure.html
(it denies deserialization if unexpected fields like `result` exist)

This PR adds `omitempty` in order to align with the general expectations.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -28,7 +28,7 @@ func NewJsonRpcRequest1(id interface{}, method string, param interface{}) *JsonR
 
 type JsonRpcResponse struct {
 	Id      interface{}     `json:"id"`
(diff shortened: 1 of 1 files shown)


metachris:
Great, thanks! 👍 

And yeah that's correct - here's the JSON-RPC docs: https://www.jsonrpc.org/specification#response_object

metachris:
Will post here once deployed (later today)

metachris:
Deployed to prod

PR #58, opened 2022-02-11
change POST to DELETE for bundle deletion

zeroXbrock:
POST seems kinda misleading, DELETE is intuitive


metachris:
Thanks 👍

metachris:
Will deploy to prod early next week

PR #59, opened 2022-02-11
filter out duplicate bundle txs

zeroXbrock:
MetaMask will repeatedly send the same transaction as time passes, so this is a measure to prevent a cached bundle from getting maxed out (16 txs) for no reason.

Not sure if the `Contains` function is the most optimal implementation -- Go isn't my top language. But it works :)


metachris:
Cool, thanks 👍 
I've cleaned it up a little in #60 - could you confirm?

PR #60, opened 2022-02-12
avoid duplicate tx in whitehat collection

metachris:
replaces #59

### DIFF
--- a/server/redisstate.go
+++ b/server/redisstate.go
@@ -194,8 +194,18 @@ func (s *RedisState) GetSenderOfTxHash(txHash string) (txSender string, found bo
 func (s *RedisState) AddTxToWhitehatBundle(bundleId string, signedTx string) error {
 	key := RedisKeyWhitehatBundleTransactions(bundleId)
 
+	// Check if item already exists
+	txs, err := s.GetWhitehatBundleTx(bundleId)
+	if err == nil {
+		for _, tx := range txs {
+			if signedTx == tx {
+				return nil
+			}
+		}
+	}
+
 	// Add item
-	err := s.RedisClient.LPush(context.Background(), key, signedTx).Err()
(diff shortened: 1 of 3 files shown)


server/redisstate.go
```diff
@@ -194,8 +194,20 @@ func (s *RedisState) GetSenderOfTxHash(txHash string) (txSender string, found bo
 func (s *RedisState) AddTxToWhitehatBundle(bundleId string, signedTx string) error {
 	key := RedisKeyWhitehatBundleTransactions(bundleId)
 
+	// Check if item already exists
+	txs, err := s.GetWhitehatBundleTx(bundleId)
+	if err != nil {
+		return err
```
bhakiyakalimuthu, line 200: do we want to fail here ? Because if there are some reason duplicate check failed and the tx is not duplicate,we might loose storing the valid tx. what do you think ?

server/redisstate.go
```diff
@@ -194,8 +194,20 @@ func (s *RedisState) GetSenderOfTxHash(txHash string) (txSender string, found bo
...
+	}
+
+	for _, tx := range txs {
+		if signedTx == tx {
+			return nil
```
bhakiyakalimuthu, line 205: good to log the duplicate tx

metachris: i think logging this will provide no benefit and just spam the logs.

server/redisstate.go
metachris: yeah good point!

zeroXbrock (approved):
looks great Chris! Thanks for the improvements

server/redisstate.go
metachris: In particular because this is an expected case, that does not signal any kind of issue or problem.

PR #62, opened 2022-02-17
[WIP] Protect analytics

bhakiyakalimuthu:
To capture data points for analytics reason, requestRecord is added.

- [x] RequestRecord,db entry creation
- [x] Store data points for each request (Right now data points are logged)
- [ ] ~~Error enum mapping~~ (Hold - handled in future)
- [x] store url & query param part of RequestEntry
- [x] add postgres store and memstore
- [x] update e2e tests
- [x] db store batch request
- [ ] ~~pgxpool support~~
- [x] modify and validate postgres sql db type to Redshift datatype


server/request_handler.go
```diff
@@ -119,9 +125,10 @@ func (r *RpcRequestHandler) processBatchRequest(client RPCProxyClient, jsonBatch
 		// Scatter worker
 		go func(count int, rpcReq *types.JsonRpcRequest) {
 			// Create child logger
-			logger := log.New(log.Ctx{"uid": fmt.Sprintf("%s.%d", r.uid, count)})
+			id := uuid.NewMD5(r.uid, big.NewInt(int64(count)).Bytes())
```
metachris, line 128: why do this?

server/request_handler.go
```diff
@@ -137,3 +144,14 @@ func (r *RpcRequestHandler) processBatchRequest(client RPCProxyClient, jsonBatch
 	// Write consolidated response
 	r._writeRpcBatchResponse(response)
 }
+
+func (r *RpcRequestHandler) requestRecord() {
+	timeRequestNeeded := time.Since(r.timeStarted) // At end of request, log the time it needed
+	r.reqRecord.requestEntry.RequestDuration = timeRequestNeeded
+	go func() {
```
metachris, line 151: this does maybe not need to happen in a goroutine

server/request_record.go
```diff
@@ -0,0 +1,48 @@
...
+type RequestRecord struct {
+	logger            log.Logger
+	requestEntry      *types.RequestEntry
+	ethSendRawTxEntry *types.EthSendRawTxEntry
+}
+
+func NewRequestRecord() *RequestRecord {
+	// If the incoming request exited in the server then below id will be used for RequestEntry
+	// If not, below id will be replaced by uid if the request forwarded to request handler
+	id := uuid.New()
```
metachris, line 21: id will be set by the database, not needed here

server/request_record.go
```diff
@@ -0,0 +1,48 @@
...
+	return &RequestRecord{
+		logger:            log.New(),
+		requestEntry:      &types.RequestEntry{Id: id},
+		ethSendRawTxEntry: &types.EthSendRawTxEntry{},
+	}
+}
+
+func (r *RequestRecord) SaveRequestEntryToDB(ctx context.Context) {
+	r.requestEntry.InsertedAt = time.Now()
```
metachris, line 30: insertedAt will be set by the database

server/request_record.go
```diff
@@ -0,0 +1,48 @@
...
+	// If not, below id will be replaced by uid if the request forwarded to request handler
+	id := uuid.New()
+	return &RequestRecord{
+		logger:            log.New(),
+		requestEntry:      &types.RequestEntry{Id: id},
+		ethSendRawTxEntry: &types.EthSendRawTxEntry{},
+	}
+}
+
+func (r *RequestRecord) SaveRequestEntryToDB(ctx context.Context) {
```
metachris, line 29: simplify to `SaveToDB`

server/request_sendrawtx.go
```diff
@@ -9,6 +9,10 @@ import (
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 )
 
+const (
+	signatureMethodByteLen = 10
```
metachris, line 13: signature is 4 bytes

server/request_sendrawtx.go
```diff
@@ -51,6 +56,14 @@ func (r *RpcRequest) handle_sendRawTransaction() {
 	r.logger.Info("[sendRawTransaction] sending raw transaction", "tx", r.tx.Hash(), "fromAddress", r.txFrom, "toAddress", AddressPtrToStr(r.tx.To()), "txNonce", r.tx.Nonce(), "txGasPrice", BigIntPtrToStr(r.tx.GasPrice()))
 	txFromLower := strings.ToLower(r.txFrom)
 
+	r.reqRecord.ethSendRawTxEntry.TxFrom = r.txFrom
+	r.reqRecord.ethSendRawTxEntry.TxTo = r.tx.To().String()
+	r.reqRecord.ethSendRawTxEntry.TxNonce = int(r.tx.Nonce())
+	r.reqRecord.ethSendRawTxEntry.TxData = r.tx.Data()
+	if len(r.tx.Data()) > signatureMethodByteLen {
+		r.reqRecord.ethSendRawTxEntry.TxSmartContractMethod = r.tx.Data()[:10]
```
metachris, line 64: use the const instead of 10 manually (`[:10]`)

types/types.go
```diff
@@ -103,3 +104,41 @@ type BundleResponse struct {
...
+	Id                 uuid.UUID     `db:"id"`
+	ReceivedAt         time.Time     `db:"received_at"`
+	InsertedAt         time.Time     `db:"inserted_at"`
+	RequestDuration    time.Duration `db:"request_duration"`
+	RequestType        string        `db:"request_type"`
+	HttpMethod         string        `db:"http_method"`
+	HttpUrl            string        `db:"http_url"`
+	HttpQueryParam     string        `db:"http_query_param"`
+	HttpResponseStatus int           `db:"http_response_status"`
+	Ip                 string        `db:"ip"`
```
metachris, line 119: should be IpHash

metachris:
good start! let's review in person tomorrow :)

server/request_handler.go
```diff
@@ -137,3 +144,14 @@ func (r *RpcRequestHandler) processBatchRequest(client RPCProxyClient, jsonBatch
...
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
```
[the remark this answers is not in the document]
bhakiyakalimuthu: Agreed

server/request_handler.go
```diff
@@ -137,3 +144,14 @@ func (r *RpcRequestHandler) processBatchRequest(client RPCProxyClient, jsonBatch
 	// Write consolidated response
 	r._writeRpcBatchResponse(response)
 }
+
+func (r *RpcRequestHandler) requestRecord() {
```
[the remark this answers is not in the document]
bhakiyakalimuthu: Updated

server/request_record.go
bhakiyakalimuthu: Yes.added this just for logging. will be updated while adding db integration

server/request_sendrawtx.go
bhakiyakalimuthu: updated the naming.

server/request_sendrawtx.go
bhakiyakalimuthu: updated

types/types.go
bhakiyakalimuthu: yes.done

server/request_record.go
bhakiyakalimuthu: Like to keep the id field same as uid so that it will make the log comparision easier during debugging.

server/request_sendrawtx.go
metachris: `tx.Data` is a byte slice, the method name should be the first 4 bytes, not 10?

server/request_record.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/flashbots/rpc-endpoint/types"
+	"github.com/google/uuid"
+	"net/http"
+	"time"
+)
+
+type RequestRecord struct {
+	logger            log.Logger
+	requestEntry      *types.RequestEntry
+	ethSendRawTxEntry *types.EthSendRawTxEntry
```
metachris, line 14: this should be a list, and the request allocate it if needed only

server/request_record.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func NewRequestRecord() *RequestRecord {
+	// If the incoming request exited in the server then below id will be used for RequestEntry
+	// If not, below id will be replaced by uid if the request forwarded to request handler
+	id := uuid.New()
+	return &RequestRecord{
+		logger:            log.New(),
+		requestEntry:      &types.RequestEntry{Id: id},
+		ethSendRawTxEntry: &types.EthSendRawTxEntry{},
```
metachris, line 24: do not initialize `ethSendRawTxEntries` here, should be done on-demand

server/request_record.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+}
+
+func (r *RequestRecord) SaveRequestEntryToDB() {
+	r.requestEntry.InsertedAt = time.Now() // this will be updated while adding db integration
+	r.logger.Info("[RequestRecord] SaveRequestEntryToDB called", "requestEntry", r.requestEntry)
+}
+func (r *RequestRecord) SaveEthSendRawTxEntryToDB() {
+	r.logger.Info("[RequestRecord] SaveEthSendRawTxEntryToDB called", "EthSendRawTxEntry", r.ethSendRawTxEntry)
+}
```
metachris, line 34: remove them eventually. the database instance will have a function `SaveRequestRecord`

server/request_processor.go
```diff
@@ -389,9 +398,22 @@ func (r *RpcRequest) writeRpcResult(result interface{}) {
...
+	r.updateRequestRecord("", 0)
 	r.jsonRes = &types.JsonRpcResponse{
 		Id:      r.jsonReq.Id,
 		Version: "2.0",
 		Result:  resBytes,
 	}
 }
+
+func (r *RpcRequest) updateRequestRecord(msg string, rpcStatusCode int) {
+	r.reqRecord.requestEntry.Error = msg
```
metachris, line 410: remove those, this function should only affect the `ethSendRawTxEntry`

server/request_processor.go
```diff
@@ -389,9 +398,22 @@ func (r *RpcRequest) writeRpcResult(result interface{}) {
...
+	r.reqRecord.requestEntry.HttpResponseStatus = rpcToHttpCode(rpcStatusCode)
```
metachris, line 411: remove those, this function should only affect the ethSendRawTxEntry

server/request_processor.go
```diff
@@ -389,9 +398,22 @@ func (r *RpcRequest) writeRpcResult(result interface{}) {
...
+	if r.jsonReq.Method == "eth_sendRawTransaction" {
+		if rpcStatusCode != 0 {
+			r.reqRecord.ethSendRawTxEntry.Error = msg
+			r.reqRecord.ethSendRawTxEntry.ErrorCode = rpcStatusCode
+		}
+		r.reqRecord.SaveEthSendRawTxEntryToDB()
```
metachris, line 417: should not save here

server/request_processor.go
```diff
@@ -389,9 +398,22 @@ func (r *RpcRequest) writeRpcResult(result interface{}) {
...
 		Version: "2.0",
 		Result:  resBytes,
 	}
 }
+
+func (r *RpcRequest) updateRequestRecord(msg string, rpcStatusCode int) {
+	r.reqRecord.requestEntry.Error = msg
+	r.reqRecord.requestEntry.HttpResponseStatus = rpcToHttpCode(rpcStatusCode)
+	if r.jsonReq.Method == "eth_sendRawTransaction" {
+		if rpcStatusCode != 0 {
```
metachris, line 413: remove if

server/request_processor.go
```diff
@@ -53,6 +58,9 @@ func (r *RpcRequest) ProcessRequest() *types.JsonRpcResponse {
 
 	switch {
 	case r.jsonReq.Method == "eth_sendRawTransaction":
+		r.reqRecord.ethSendRawTxEntry.IsWhiteHatBundleCollection = r.isWhitehatBundleCollection
```
metachris, line 61: `r.reqRecord.NewEthSendRawTxEntry`, and keep the instance around (probably as part of the request)

types/types.go
```diff
@@ -103,3 +104,42 @@ type BundleResponse struct {
...
+
+// EthSendRawTxEntry to store each eth_sendRawTransaction calls
+type EthSendRawTxEntry struct {
+	Id                         uuid.UUID `db:"id"`
+	RequestId                  uuid.UUID `db:"request_id"` // id from RequestEntry table
+	IsOnOafcList               bool      `db:"is_on_oafc_list"`
+	IsWhiteHatBundleCollection bool      `db:"is_white_hat_bundle_collection"`
+	WhiteHatBundleId           string    `db:"white_hat_bundle_id"`
+	IsCancelTx                 bool      `db:"is_cancel_tx"`
+	IsTxSentToRelay            bool      `db:"is_tx_sent_to_relay"`
```
metachris, line 134: maybe we should do `NeedsFrontrunningProtection` and `WasSentToRelay`

server/request_sendrawtx.go
```diff
@@ -9,6 +9,10 @@ import (
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 )
 
+const (
+	signatureMethodLen = 10 // first 4 byte of data field
```
metachris, line 13: should be 4

server/request_handler.go
```diff
@@ -71,41 +70,46 @@ func (r *RpcRequestHandler) process() {
...
 			(*r.respw).WriteHeader(http.StatusBadRequest)
 			return
 		}
+		r.reqRecord.requestEntry.RequestType = "batch"
 		// Process batch request
 		r.processBatchRequest(client, jsonBatchReq, ip, origin, isWhitehatBundleCollection, whitehatBundleId)
 		return
 	}
-
+	r.reqRecord.requestEntry.RequestType = "single"
```
metachris, line 104: maybe should be a bool field `isBatchRequest`, and perhaps we add `numBatchRequests` 🤔

server/server.go
```diff
@@ -54,6 +57,14 @@ func NewRpcEndPointServer(version string, listenAddress, proxyUrl, relayUrl stri
 		redisUrl = redisServer.Addr()
 	}
 
+	// Setup database
+	var db database.Store
+	if psqlDsn == "" {
+		db = testutils.NewMockStore()
```
bhakiyakalimuthu, line 63: Should we keep mocks as part of testutils?

sql/001_initial.up.sql
```diff
@@ -0,0 +1,41 @@
+BEGIN;
+CREATE SCHEMA IF NOT EXISTS requests;
+CREATE TABLE requests.main(
```
bhakiyakalimuthu, line 3: like to rename `requests.main`. But not getting right naming.

Makefile
```diff
@@ -12,12 +12,16 @@ clean:
 	rm -rf rpc-endpoint build/
 
 test:
-
 	go test ./...
-test_up:
+migrate_up:
+	docker run -v ${CURDIR}/sql:/migrations --network host migrate/migrate -path=/migrations/ -database 'postgres://postgres:postgres@localhost:5432/test?sslmode=disable&search_path=requests' up
+migrate_down:
+	docker run -v ${CURDIR}/sql:/migrations --network host migrate/migrate -path=/migrations/ -database 'postgres://postgres:postgres@localhost:5432/test?sslmode=disable&search_path=requests' down -all
```
bhakiyakalimuthu, line 19: very useful for migration - https://github.com/golang-migrate/migrate
it also support different db as well as redshift support.
https://github.com/golang-migrate/migrate/tree/master/database/redshift

server/request_processor.go
```diff
@@ -53,7 +62,14 @@ func (r *RpcRequest) ProcessRequest() *types.JsonRpcResponse {
...
 	switch {
 	case r.jsonReq.Method == "eth_sendRawTransaction":
+		r.ethSendRawTxEntry = &database.EthSendRawTxEntry{
+			IsWhiteHatBundleCollection: r.isWhitehatBundleCollection,
+			WhiteHatBundleId:           r.whitehatBundleId,
+			Id:                         r.id,
+			RequestId:                  r.requestId,
+		}
 		r.handle_sendRawTransaction()
+		r.db.SaveEthSendRawTxEntry(r.ethSendRawTxEntry)
```
metachris, line 72: do not pass the database in to request, the whole thing should be saved at once by the handler.

server/request_handler.go
```diff
@@ -137,3 +144,10 @@ func (r *RpcRequestHandler) processBatchRequest(client RPCProxyClient, jsonBatch
 	// Write consolidated response
 	r._writeRpcBatchResponse(response)
 }
+
+func (r *RpcRequestHandler) finishRequest() {
+	timeRequestNeeded := time.Since(r.timeStarted) // At end of request, log the time it needed
+	r.requestEntry.RequestDuration = timeRequestNeeded
+	r.db.SaveRequestEntry(r.requestEntry)
```
metachris, line 151: the database should be able to take the whole record and save the request and each individual ethsendrawtxreq from there

bhakiyakalimuthu: I also thought about it in the beginning and This will not affect single request.But this logic will have extra looping & extra list/channel for batch request.More changes.

Pros
1.db is not passed everywhere.
2.Save db is called from a single place.

Cons
1.In the Current implementation each individual eth_sendRawRequests are saved inside the request itself and not much change required except the db is constructed during the request init phase. If we want to do it in the handler then we will have to return each eth_sendRawRequest and consolidate inside the list.Similar to r.JsonRes. Also extra list and channel needed.Method signature needs to be changed for  
`(r *RpcRequest) ProcessRequest()`
  method . 
2. During the finishRequest phase,the list needs to be passed inside the db method(SaveEthSendRawTxEntry) where we will have to loop in to store each individual eth_sendRawRequest.

metachris: i think the EthSendRawTxEntry should probably be created by the handler and the pointer passed to the request. the RequestRecord should contain everything, and the database should be able to save everything within one batched request.

Makefile
```diff
@@ -13,7 +13,16 @@ clean:
 
 test:
 	go test ./...
-
+migrate_up:
+	docker run -v ${CURDIR}/sql:/migrations --network host migrate/migrate -path=/migrations/ -database 'postgres://postgres:postgres@localhost:5432/test?sslmode=disable&search_path=requests' up
+migrate_down:
```
metachris, line 18: what's the use-case for `migrate_down`?

cmd/server/main.go
```diff
@@ -20,6 +22,8 @@ var (
 	defaultProxyUrl      = "http://127.0.0.1:8545"
 	defaultRelayUrl      = "https://relay.flashbots.net"
 	defaultRedisUrl      = "localhost:6379"
+	defaultPostgresDSN   = fmt.Sprintf("host=%s port=%d dbname=%s user=%s password=%s sslmode=disable",
```
metachris, line 25:
```suggestion
	defaultPostgresDSN   = "postgres://postgres:postgres@localhost:5432/test?sslmode=disable",
```

database/mem_store.go
```diff
@@ -0,0 +1,34 @@
...
+	"sync"
+)
+
+type memStore struct {
+	requests      map[uuid.UUID]*RequestEntry
+	ethSendRawTxs map[uuid.UUID]*EthSendRawTxEntry
+	mutex         *sync.Mutex
+}
+
+func NewMemStore(requests map[uuid.UUID]*RequestEntry, ethSendRawTxs map[uuid.UUID]*EthSendRawTxEntry) Store {
```
metachris, line 14: `NewMemStore` should not require arguments, but just create the empty maps inside

database/postgres_store.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+func (d *postgresStore) Close() {
+	d.DB.Close()
+}
+
+func (d *postgresStore) SaveRequestEntry(in *RequestEntry) error {
+	query := `INSERT INTO requests.main 
+	(id,received_at,inserted_at,request_duration,is_batch_request,num_request_in_batch,http_method,http_url,http_query_param,http_response_status,ip_hash,origin,host,error) VALUES (:id,:received_at,:inserted_at,:request_duration,:is_batch_request,:num_request_in_batch,:http_method,:http_url,:http_query_param,:http_response_status,:ip_hash,:origin,:host,:error)`
```
metachris, line 32: add spaced between fields pls

go.mod
```diff
@@ -3,16 +3,21 @@ module github.com/flashbots/rpc-endpoint
...
 )
+
+require (
+	github.com/alicebob/gopher-json v0.0.0-20200520072559-a9ecdc9d1d3a // indirect
+	github.com/gomodule/redigo v1.8.5 // indirect
+	github.com/lib/pq v1.10.4 // indirect
+	github.com/yuin/gopher-lua v0.0.0-20200816102855-ee81675732da // indirect
+    golang.org/x/crypto v0.0.0-20211117183948-ae814b36b871 // indirect
+    golang.org/x/sys v0.0.0-20211124211545-fe61309f8881 // indirect
+)
```
metachris, line 23: seems this file is a bit unclean, please re-run `go mod tidy`

server/server.go
```diff
@@ -183,3 +200,9 @@ func IsBlacklisted(ip string) bool {
 	}
 	return false
 }
+
+func setHeader(respw http.ResponseWriter) http.ResponseWriter {
```
metachris, line 204: does not need to return the writer, and should probably be called `setCorsHeaders`

tests/e2e_test.go
```diff
@@ -46,6 +48,12 @@ var bundleJsonApi *httptest.Server
 
 // Reset the RPC endpoint and mock backend servers
 func resetTestServers() {
+	redisServerAddr, rpcBackendUrl := serverPreSetup()
+	db := testutils.NewMockStore()
+	serverSetup(redisServerAddr, rpcBackendUrl, db)
+}
+
+func serverPreSetup() (string, string) {
```
metachris, line 56: why split this up (why create `serverPreSetup`)?

server/server.go
bhakiyakalimuthu: updated

go.mod
bhakiyakalimuthu: done

database/mem_store.go
bhakiyakalimuthu: updated

cmd/server/main.go
bhakiyakalimuthu: done

sql/001_initial.schema.up.sql
```diff
@@ -0,0 +1,39 @@
+BEGIN;
+CREATE TABLE rpc_endpoint_requests(
+    id uuid not null unique primary key default gen_random_uuid(),
+    received_at timestamp with time zone not null default now(),
+    inserted_at timestamp with time zone not null default now(),
+    request_duration interval,
+    is_batch_request boolean,
+    num_request_in_batch integer,
+    http_method varchar,
```
bhakiyakalimuthu, line 9: add constrains

database/postgres_store.go
```diff
@@ -0,0 +1,95 @@
...
+	defer cancel()
+	if _, err := d.DB.Exec(ctx, query, in.Id, in.ReceivedAt, time.Now(), in.RequestDuration, in.IsBatchRequest, in.NumRequestInBatch, in.HttpMethod, in.HttpUrl, in.HttpQueryParam, in.HttpResponseStatus, in.IpHash, in.Origin, in.Host, in.Error); err != nil {
+		log.Error("[postgresStore] SaveRequestEntry failed", "error", err)
+		return err
+	}
+	log.Info("[postgresStore] SaveRequestEntry succeeded", "RequestEntry", in) // TODO:Remove logging
+	return nil
+}
+
+func (d *postgresStore) SaveEthSendRawTxEntries(in []*EthSendRawTxEntry) error {
```
metachris, line 47: rename `in` to `rawTxEntries`

database/postgres_store.go
```diff
@@ -0,0 +1,95 @@
+package database
+
+import (
+	"context"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/jackc/pgx/v4"
```
metachris, line 6: please use https://github.com/jmoiron/sqlx/ (sqlx is used in all our other projects too)

cmd/server/main.go
```diff
@@ -28,6 +29,7 @@ var (
 	redisUrl        = flag.String("redis", getEnvOrDefault("REDIS_URL", defaultRedisUrl), "URL for Redis (use 'dev' to use integrated in-memory redis)")
 	relayUrl        = flag.String("relayUrl", getEnvOrDefault("RELAY_URL", defaultRelayUrl), "URL for relay")
 	relaySigningKey = flag.String("signingKey", os.Getenv("RELAY_SIGNING_KEY"), "Signing key for relay requests")
+	psqlDsn         = flag.String("psqlDsn", os.Getenv("POSTGRES_DSN"), "Postgres DSN")
```
metachris, line 32:
```suggestion
	psqlDsn         = flag.String("psql", os.Getenv("POSTGRES_DSN"), "Postgres DSN")
```

database/postgres_store.go
```diff
@@ -0,0 +1,53 @@
...
+	return &postgresStore{
+		DB: db,
+	}
+}
+
+func (d *postgresStore) Close() {
+	d.DB.Close()
+}
+
+func (d *postgresStore) SaveRequestEntry(in *RequestEntry) error {
```
metachris, line 30: rename `in` to `entry`

database/postgres_store.go
```diff
@@ -0,0 +1,53 @@
...
+	defer cancel()
+	if _, err := d.DB.NamedExecContext(ctx, query, in); err != nil {
+		log.Error("[postgresStore] SaveRequestEntry failed", "id", in.Id, "error", err)
+		return err
+	}
+	log.Info("[postgresStore] SaveRequestEntry succeeded", "RequestEntry", in) // TODO:Remove logging
+	return nil
+}
+
+func (d *postgresStore) SaveRawTxEntries(in []*EthSendRawTxEntry) error {
```
metachris, line 44: rename `in` to `entries`

database/postgres_store.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (d *postgresStore) SaveRequestEntry(in *RequestEntry) error {
+	in.InsertedAt = time.Now()
+	query := `INSERT INTO rpc_endpoint_requests 
+	(id, received_at, inserted_at, request_duration_ms, is_batch_request, num_request_in_batch, http_method, http_url, http_query_param, http_response_status, ip_hash, origin, host, error) VALUES (:id, :received_at, :inserted_at, :request_duration_ms, :is_batch_request, :num_request_in_batch, :http_method, :http_url, :http_query_param, :http_response_status, :ip_hash, :origin, :host, :error)`
+	ctx, cancel := context.WithTimeout(context.Background(), connTimeOut)
+	defer cancel()
+	if _, err := d.DB.NamedExecContext(ctx, query, in); err != nil {
+		log.Error("[postgresStore] SaveRequestEntry failed", "id", in.Id, "error", err)
```
metachris, line 37: logging should be done where this function is called

database/postgres_store.go
```diff
@@ -0,0 +1,53 @@
...
+		return err
+	}
+	log.Info("[postgresStore] SaveRequestEntry succeeded", "RequestEntry", in) // TODO:Remove logging
```
metachris, line 40: logging should be done outside of db

server/request_handler.go
```diff
@@ -21,36 +20,38 @@ type RpcRequestHandler struct {
...
 	defaultProxyUrl string
 	relaySigningKey *ecdsa.PrivateKey
-	uid             string
+	uid             uuid.UUID
+	db              database.Store
+	requestRecord   *requestRecord
 }
 
-func NewRpcRequestHandler(respw *http.ResponseWriter, req *http.Request, proxyUrl string, relaySigningKey *ecdsa.PrivateKey) *RpcRequestHandler {
+func NewRpcRequestHandler(respw *http.ResponseWriter, req *http.Request, proxyUrl string, relaySigningKey *ecdsa.PrivateKey, db database.Store, requestRecord *requestRecord) *RpcRequestHandler {
```
metachris, line 28: requestRecord should be created here, not passed in from the outside?

server/request_handler.go
```diff
@@ -117,14 +126,17 @@ func (r *RpcRequestHandler) processBatchRequest(client RPCProxyClient, jsonBatch
...
 		// Process each individual request
 		// Scatter worker
-		go func(count int, rpcReq *types.JsonRpcRequest) {
+		go func(count int, rpcReq *types.JsonRpcRequest, record *requestRecord) {
+			id := uuid.New()
 			// Create child logger
-			logger := log.New(log.Ctx{"uid": fmt.Sprintf("%s.%d", r.uid, count)})
+			logger := log.New(log.Ctx{"uid": r.uid, "id": id, "count": count})
+
+			entry := r.requestRecord.UpdateEthSendRawTxEntries(rpcReq)
```
metachris, line 134: in there, the `RequestId` should already be set, then you don't need to pass it into `NewRpcRequest`

metachris: Also I think `UpdateEthSendRawTxEntries` does too much magic. I think more easily readable would be to have the check if this method is eth_sendRawTx here on the outside, and rename `UpdateEthSendRawTxEntries` to `AddEthSendRawTxEntry`

server/request_processor.go
```diff
@@ -33,18 +36,25 @@ type RpcRequest struct {
 	origin                     string
 	isWhitehatBundleCollection bool
 	whitehatBundleId           string
+	id                         uuid.UUID
+	requestId                  uuid.UUID
```
metachris, line 40: not needed here, should be already set on the record entry

tests/e2e_test.go
```diff
@@ -44,8 +45,13 @@ func init() {
 
 var bundleJsonApi *httptest.Server
 
-// Reset the RPC endpoint and mock backend servers
-func resetTestServers() {
+// Setup RPC endpoint and mock backend servers
+func testSetup() {
+	db := database.NewMockStore()
+	serverSetup(db)
+}
```
metachris, line 52: why split this into 2 functions?

database/types.go
```diff
@@ -0,0 +1,48 @@
...
+type EthSendRawTxEntry struct {
+	Id                          uuid.UUID `db:"id"`
+	RequestId                   uuid.UUID `db:"request_id"` // id from RequestEntry table
+	InsertedAt                  time.Time `db:"inserted_at"`
+	IsOnOafcList                bool      `db:"is_on_oafc_list"`
+	IsWhiteHatBundleCollection  bool      `db:"is_white_hat_bundle_collection"`
+	WhiteHatBundleId            string    `db:"white_hat_bundle_id"`
+	IsCancelTx                  bool      `db:"is_cancel_tx"`
+	NeedsFrontRunningProtection bool      `db:"needs_front_running_protection"`
+	WasSentToRelay              bool      `db:"was_sent_to_relay"`
```
metachris, line 36: I find it confusing to have `WasSentToRelay` and `IsTxSentToRelay` since it means the same thing. I think better would be:

* `ShouldSendToRelay`
* `WasSentToRelay`

server/request_processor.go
```diff
@@ -33,18 +36,25 @@ type RpcRequest struct {
 	origin                     string
 	isWhitehatBundleCollection bool
 	whitehatBundleId           string
+	id                         uuid.UUID
```
metachris, line 39: not needed here, should be already set on the record entry

server/request_response.go
```diff
@@ -50,3 +40,18 @@ func (r *RpcRequestHandler) _writeRpcBatchResponse(res []*types.JsonRpcResponse)
 		(*r.respw).WriteHeader(http.StatusInternalServerError)
 	}
 }
+
+func rpcToHttpCode(rpcStatusCode int) int {
```
metachris, line 44: Where did you find this mapping? Shouldn't http status code be 200 even on errors, using JSON-RPC?

server/request_sendrawtx.go
```diff
@@ -90,6 +107,7 @@ func (r *RpcRequest) handle_sendRawTransaction() {
 	// Check for cancellation-tx
 	if len(r.tx.Data()) <= 2 && txFromLower == strings.ToLower(r.tx.To().Hex()) {
 		requestDone := r.handleCancelTx() // returns true if tx was cancelled at the relay and response has been sent to the user
+		r.ethSendRawTxEntry.IsCancelTx = requestDone
```
metachris, line 110:
```suggestion
		r.ethSendRawTxEntry.IsCancelTx = true
```

This is a cancel-tx already, no matter if it has been sent to the relay or not.

server/util.go
```diff
@@ -94,6 +95,15 @@ func GetIP(r *http.Request) string {
 	return r.RemoteAddr
 }
 
+func GetIPHash(r *http.Request) string {
+	ipHash := md5.New()
+	_, err := ipHash.Write([]byte(GetIP(r)))
+	if err != nil {
+		return ""
+	}
+	return hex.EncodeToString(ipHash.Sum(nil))
```
metachris, line 104: wouldn't it be simpler like this:

```go
   hash := md5.Sum([]byte(GetIP(r)))
   return hex.EncodeToString(hash[:])
```

server/request_handler.go
```diff
@@ -137,3 +149,13 @@ func (r *RpcRequestHandler) processBatchRequest(client RPCProxyClient, jsonBatch
...
 }
+
+func (r *RpcRequestHandler) finishRequest() {
+	timeRequestNeeded := time.Since(r.timeStarted) // At end of request, log the time it needed
+	r.requestRecord.requestEntry.RequestDurationMs = timeRequestNeeded.Milliseconds()
+	if len(r.requestRecord.ethSendRawTxEntries) > 0 {
+		r.db.SaveRequestEntry(r.requestRecord.requestEntry)
+		r.db.SaveRawTxEntries(r.requestRecord.ethSendRawTxEntries)
+	}
+	r.logger.Info("Request finished", "timeTakenInSec", timeRequestNeeded.Seconds())
```
metachris, line 160:
```suggestion
	r.logger.Info("Request finished", "duration", timeRequestNeeded.Seconds())
```

server/request_handler.go
```diff
@@ -137,3 +149,13 @@ func (r *RpcRequestHandler) processBatchRequest(client RPCProxyClient, jsonBatch
 	// Write consolidated response
 	r._writeRpcBatchResponse(response)
 }
+
+func (r *RpcRequestHandler) finishRequest() {
+	timeRequestNeeded := time.Since(r.timeStarted) // At end of request, log the time it needed
+	r.requestRecord.requestEntry.RequestDurationMs = timeRequestNeeded.Milliseconds()
+	if len(r.requestRecord.ethSendRawTxEntries) > 0 {
+		r.db.SaveRequestEntry(r.requestRecord.requestEntry)
```
metachris, line 157: I think there should be 1 call `r.db.SaveRequest` that saves both the requestEntry as well as all the rawTxEntries, and that should be spun off into a goroutine to not block the main request.

bhakiyakalimuthu: Done

server/request_handler.go
bhakiyakalimuthu: Done

server/util.go
bhakiyakalimuthu: updated

server/request_sendrawtx.go
bhakiyakalimuthu: updated

server/request_response.go
bhakiyakalimuthu: yes it works as what you mentioned..we respond with 200 ok for errors. this is used in requestResponse

server/request_processor.go
bhakiyakalimuthu: updated

server/request_processor.go
bhakiyakalimuthu: updated

server/request_handler.go
bhakiyakalimuthu: updated

server/request_handler.go
bhakiyakalimuthu: updated

database/postgres_store.go
bhakiyakalimuthu: updated. Instead of logging inside each method, this will be logged from SaveRequest method

database/postgres_store.go
bhakiyakalimuthu: updated

database/postgres_store.go
bhakiyakalimuthu: done

bhakiyakalimuthu: Updated

database/postgres_store.go
bhakiyakalimuthu: Updated

database/types.go
bhakiyakalimuthu: Updated

tests/e2e_test.go
bhakiyakalimuthu: Because depends on the test, we need to pass in mem store or mock store..If you have someother ideas here, let me know.I can update it accordingly.

cmd/server/main.go
bhakiyakalimuthu: updated

server/request_processor.go
```diff
@@ -163,7 +168,9 @@ func (r *RpcRequest) sendTxToRelay() {
 	txHash := strings.ToLower(r.tx.Hash().Hex())
 
 	// Check if tx was already forwarded and should be blocked now
-	if r.blockResendingTxToRelay(txHash) {
+	wasSentToRelay := r.blockResendingTxToRelay(txHash)
+	r.ethSendRawTxEntry.WasSentToRelay = wasSentToRelay
```
metachris, line 172: here, `ShouldSendToRelay` should be set to true

server/request_processor.go
```diff
@@ -176,7 +183,7 @@ func (r *RpcRequest) sendTxToRelay() {
 	if err != nil {
 		r.logger.Error("[sendTxToRelay] Redis:SetTxSentToRelay failed", "error", err)
 	}
-
+	r.ethSendRawTxEntry.ShouldSendToRelay = true
```
metachris, line 186: wrong place here, move to top of function

server/request_record.go
```diff
@@ -0,0 +1,49 @@
...
+	return &requestRecord{
+		requestEntry:        &database.RequestEntry{},
+		ethSendRawTxEntries: make([]*database.EthSendRawTxEntry, 0),
+		mutex:               &sync.Mutex{},
+	}
+}
+
+func (r *requestRecord) AddEthSendRawTxEntry(jsonReq *types.JsonRpcRequest, id, requestId uuid.UUID) *database.EthSendRawTxEntry {
+	var entry *database.EthSendRawTxEntry
+	if jsonReq.Method == "eth_sendRawTransaction" {
```
metachris, line 27: i'm not a fan that this function does or not does it, depending on jsonReq. I think this function should ideally not take jsonReq at all, and maybe the RpcRequest should call this at the correct spot? 🤔

tests/e2e_test.go
metachris: maybe rename `testSetup` to `serverSetupWithMockStore`

server/request_processor.go
bhakiyakalimuthu: That means `NeedsFrontRunningProtection = ShouldSendToRelay ` ?

metachris: oh right, we have that already! good point. could remove `ShouldSendToRelay`? 🤔

server/request_record.go
bhakiyakalimuthu: Done

server/request_processor.go
bhakiyakalimuthu: updated

server/request_processor.go
bhakiyakalimuthu: Removed `ShouldSendToRelay`

sql/redshift/001_initial.schema.up.sql
```diff
@@ -0,0 +1,39 @@
+BEGIN;
+CREATE TABLE rpc_endpoint_requests(
+    id varchar(128) not null distkey,
+    received_at timestamptz default sysdate,
```
metachris, line 4: should this really have `default sysdate` / isn't this supposed to be set by the application?

server/request_record.go
```diff
@@ -0,0 +1,60 @@
...
+		RequestId: requestId,
+	}
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	r.ethSendRawTxEntries = append(r.ethSendRawTxEntries, entry)
+	return entry
+}
+
+func (r *requestRecord) UpdateRequestEntry(req *http.Request, reqStatus int, error string) {
+	// TODO:Error should be converted to enum
```
metachris, line 39: remove the comment?

server/request_record.go
```diff
@@ -0,0 +1,60 @@
...
+	r.requestEntry.IpHash = GetIPHash(req)
+	r.requestEntry.Error = error
+	r.requestEntry.HttpUrl = req.URL.Path
+	r.requestEntry.HttpQueryParam = req.URL.RawQuery
+	r.requestEntry.HttpResponseStatus = reqStatus
+	r.requestEntry.Origin = req.Header.Get("Origin")
+	r.requestEntry.Host = req.Header.Get("Host")
+}
+
+func (r *requestRecord) SaveRecord() {
```
metachris, line 50: error should be returned, and logging should perhaps be done outside?

bhakiyakalimuthu: Yes error can be returned but it will be omitted on the caller side. Its kind of hard to log on the caller side.

sql/redshift/001_initial.schema.up.sql
```diff
@@ -0,0 +1,39 @@
...
+    inserted_at timestamptz default sysdate sortkey,
+    request_duration_ms bigint not null,
+    is_batch_request boolean,
+    num_request_in_batch integer,
+    http_method varchar(10),
+    http_url varchar(20),
```
metachris, line 10: perhaps change to 100?

server/request_sendrawtx.go
```diff
@@ -51,6 +57,16 @@ func (r *RpcRequest) handle_sendRawTransaction() {
...
 	txFromLower := strings.ToLower(r.txFrom)
 
+	// store tx info to ethSendRawTxEntries which will be stored in db for data analytics reason
+	r.ethSendRawTxEntry.TxFrom = r.txFrom
+	r.ethSendRawTxEntry.TxTo = AddressPtrToStr(r.tx.To())
+	r.ethSendRawTxEntry.TxNonce = int(r.tx.Nonce())
+	r.ethSendRawTxEntry.TxData = common.Bytes2Hex(r.tx.Data())
+
+	if len(r.tx.Data()) >= signatureMethodByteLen {
+		r.ethSendRawTxEntry.TxSmartContractMethod = common.Bytes2Hex(r.tx.Data()[:signatureMethodByteLen])
```
metachris, line 67: add 0x prefix, as it's custom

server/request_sendrawtx.go
```diff
@@ -3,12 +3,17 @@ package server
...
 	"fmt"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/flashbots/rpc-endpoint/types"
 	"strings"
 
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 )
 
+const (
+	signatureMethodByteLen = 4 // first 4 byte of data field
```
metachris, line 14: rename to `scMethodBytes`

sql/redshift/001_initial.schema.up.sql
bhakiyakalimuthu: yes it is. but default sytemdate.

server/request_record.go
bhakiyakalimuthu: updated

server/request_record.go
bhakiyakalimuthu: updated

sql/redshift/001_initial.schema.up.sql
bhakiyakalimuthu: updated

sql/redshift/001_initial.schema.up.sql
```diff
@@ -0,0 +1,39 @@
...
+    is_blocked_bcz_already_sent boolean,
+    error varchar(1000),
+    error_code integer,
+    tx_raw varchar(max),
+    tx_hash varchar(66),
+    tx_from varchar(42),
+    tx_to varchar(42),
+    tx_nonce integer,
+    tx_data varchar(10000),
+    tx_smart_contract_method varchar(8)
```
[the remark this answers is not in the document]
bhakiyakalimuthu: done

server/request_sendrawtx.go
bhakiyakalimuthu: done

server/request_sendrawtx.go
bhakiyakalimuthu: done

sql/redshift/001_initial.schema.up.sql
```diff
@@ -0,0 +1,40 @@
+BEGIN;
+CREATE TABLE rpc_endpoint_requests(
+    id varchar(128) not null distkey,
+    received_at timestamptz not null default sysdate,
```
metachris, line 4: not sure why, but my comment here disappeared: https://github.com/flashbots/rpc-endpoint/pull/62/files#diff-49621259d00261dd2798c920f922ceb9411562f3ef333791864b480f61a32446R4 -- received_at should not use default sysdate because it should be set by the application

sql/psql/001_initial.schema.up.sql
```diff
@@ -0,0 +1,39 @@
+BEGIN;
+CREATE TABLE rpc_endpoint_requests(
+    id uuid not null unique primary key default gen_random_uuid(),
+    received_at timestamp with time zone not null default now(),
```
metachris, line 4: should not use default because must be set by application

server/request_record.go
```diff
@@ -0,0 +1,71 @@
...
+		if err := r.db.SaveRawTxEntries(r.ethSendRawTxEntries); err != nil {
+			return fmt.Errorf("SaveRawTxEntries failed %v", err)
+		}
+	}
+	return nil
+}
+
+// getForwardedRawTxEntries returns list of rawTxEntry which are either sent to relay or mempool or entry with error
+func (r *requestRecord) getForwardedRawTxEntries() []*database.EthSendRawTxEntry {
+	entries := make([]*database.EthSendRawTxEntry, 0, len(r.ethSendRawTxEntries))
```
metachris, line 64:
```suggestion
	entries := make([]*database.EthSendRawTxEntry)
```

metachris, line 64:
```suggestion
func (r *requestRecord) getForwardedRawTxEntries() (entries []*database.EthSendRawTxEntry) {
```

metachris: maybe we should rename this method to `getRawTxEntriesToSave`, because it's not only forwarded ones? 🤔

server/request_record.go
```diff
@@ -0,0 +1,71 @@
...
+	if len(r.getForwardedRawTxEntries()) > 0 { // Save entries if the request contains rawTxEntries
+		if err := r.db.SaveRequestEntry(r.requestEntry); err != nil {
+			return fmt.Errorf("SaveRequestEntry failed %v", err)
+		}
+		if err := r.db.SaveRawTxEntries(r.ethSendRawTxEntries); err != nil {
+			return fmt.Errorf("SaveRawTxEntries failed %v", err)
+		}
+	}
+	return nil
+}
```
metachris, line 60:
```suggestion
func (r *requestRecord) SaveRecord() error {
	entries := r.getForwardedRawTxEntries()
	if len(entries) > 0 {
		if err := r.db.SaveRequestEntry(r.requestEntry); err != nil {
			return fmt.Errorf("SaveRequestEntry failed %v", err)
		}
		if err := r.db.SaveRawTxEntries(entries); err != nil {
			return fmt.Errorf("SaveRawTxEntries failed %v", err)
		}
	}
	return nil
}
```

server/request_record.go
```diff
@@ -0,0 +1,71 @@
...
+
+// getForwardedRawTxEntries returns list of rawTxEntry which are either sent to relay or mempool or entry with error
+func (r *requestRecord) getForwardedRawTxEntries() []*database.EthSendRawTxEntry {
+	entries := make([]*database.EthSendRawTxEntry, 0, len(r.ethSendRawTxEntries))
+	for _, entry := range r.ethSendRawTxEntries {
```
metachris, line 65: should probably use the lock here too

server/request_record.go
```diff
@@ -0,0 +1,71 @@
...
+	requestEntry        *database.RequestEntry
+	ethSendRawTxEntries []*database.EthSendRawTxEntry
+	mutex               *sync.Mutex
+	db                  database.Store
+}
+
+func NewRequestRecord(db database.Store) *requestRecord {
+	return &requestRecord{
+		requestEntry:        &database.RequestEntry{},
+		ethSendRawTxEntries: make([]*database.EthSendRawTxEntry, 0),
```
metachris, line 21:
```suggestion
		ethSendRawTxEntries: make([]*database.EthSendRawTxEntry),
```

server/request_record.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/google/uuid"
+	"net/http"
+	"sync"
+)
+
+type requestRecord struct {
+	requestEntry        *database.RequestEntry
+	ethSendRawTxEntries []*database.EthSendRawTxEntry
+	mutex               *sync.Mutex
+	db                  database.Store
```
metachris, line 15: shouldn't this be a pointer to the db store?

server/request_record.go
bhakiyakalimuthu: good catch :-)

server/request_record.go
bhakiyakalimuthu: Its an interface. Not sure what do you mean ?

metachris: How is the actual database instance passed in here, by value or by reference?

server/request_record.go
```diff
@@ -0,0 +1,71 @@
...
+		ethSendRawTxEntries: make([]*database.EthSendRawTxEntry, 0),
+		mutex:               &sync.Mutex{},
+		db:                  db,
+	}
+}
+
+func (r *requestRecord) AddEthSendRawTxEntry(id, requestId uuid.UUID) *database.EthSendRawTxEntry {
+	entry := &database.EthSendRawTxEntry{
+		Id:        id,
+		RequestId: requestId,
```
metachris, line 30: shouldn't it be able to take requestId from `r.requestEntry`?

server/request_record.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"fmt"
+	"github.com/flashbots/rpc-endpoint/database"
+	"github.com/google/uuid"
+	"net/http"
+	"sync"
+)
+
+type requestRecord struct {
+	requestEntry        *database.RequestEntry
```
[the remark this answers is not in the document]
bhakiyakalimuthu: updated

server/request_record.go
```diff
@@ -0,0 +1,71 @@
...
+	ethSendRawTxEntries []*database.EthSendRawTxEntry
+	mutex               *sync.Mutex
```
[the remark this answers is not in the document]
bhakiyakalimuthu: updated

server/request_record.go
bhakiyakalimuthu: updated

server/request_record.go
bhakiyakalimuthu: updated

sql/redshift/001_initial.schema.up.sql
bhakiyakalimuthu: updated

sql/psql/001_initial.schema.up.sql
bhakiyakalimuthu: updated

server/request_record.go
metachris: i mean why wouldn't you replace this with `RequestId: r.RequestEntry.Id`?

server/request_record.go
bhakiyakalimuthu: by reference

server/request_record.go
bhakiyakalimuthu: updated

server/request_record.go
```diff
@@ -0,0 +1,74 @@
...
+	ethSendRawTxEntries []*database.EthSendRawTxEntry
+	mutex               sync.Mutex
+	db                  database.Store
+}
+
+func NewRequestRecord(db database.Store) *requestRecord {
+	return &requestRecord{
+		requestEntry:        database.RequestEntry{},
+		ethSendRawTxEntries: make([]*database.EthSendRawTxEntry, 0),
+		mutex:               sync.Mutex{},
```
metachris, line 22:
```suggestion
```

metachris: You can remove all those lines

bhakiyakalimuthu: done

server/request_record.go
metachris: still todo

database/postgres_store.go
```diff
@@ -0,0 +1,45 @@
...
+		DB: db,
+	}
+}
+
+func (d *postgresStore) Close() {
+	d.DB.Close()
+}
+
+func (d *postgresStore) SaveRequestEntry(entry RequestEntry) error {
+	entry.InsertedAt = time.Now()
```
metachris, line 30: should be set by database

Issue #66, opened 2022-03-04
Support contract creation


zzh1996:
Sending contract creation tx (to=null) to Flashbots RPC will return error now

metachris:
deployed

PR #67, opened 2022-03-04
allow transactions to null target (contract creation)

metachris:
See #66

### DIFF
--- a/server/request_processor.go
+++ b/server/request_processor.go
@@ -177,12 +177,6 @@ func (r *RpcRequest) sendTxToRelay() {
 		r.logger.Error("[sendTxToRelay] Redis:SetTxSentToRelay failed", "error", err)
 	}
 
-	txTo := r.tx.To()
-	if txTo == nil {
-		r.writeRpcError("invalid target", types.JsonRpcInternalError)
-		return
-	}
-
 	minNonce, maxNonce, err := r.GetAddressNonceRange(r.txFrom)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


bertmiller:
lgtm on a quick review, will wait for @bhakiyakalimuthu to merge

server/request_processor.go
```diff
@@ -199,12 +193,16 @@ func (r *RpcRequest) sendTxToRelay() {
 	// only allow large transactions to certain addresses - default max tx size is 128KB
 	// https://github.com/ethereum/go-ethereum/blob/master/core/tx_pool.go#L53
 	if r.tx.Size() > 131072 {
-		if _, found := allowedLargeTxTargets[txTo.Hex()]; !found {
-			r.logger.Error("[sendTxToRelay] large tx not allowed to target", "tx", txHash, "target", txTo)
+		if r.tx.To() == nil {
```
bhakiyakalimuthu, line 196: @metachris For larger tx with nil To address, we still set it as sentToRelay (True) in redis even though this tx is not sent to redis. I assume we are ok with that.

bhakiyakalimuthu (approved):
LGTM :-)

server/request_processor.go
metachris: Yes, that's fine just as before, it's just used for spam protection - to prevent resending the same tx over and over. If this tx would be sent again it would run into the same error, so that's fine.

PR #70, opened 2022-03-16
fast endpoint + preferences

bhakiyakalimuthu:
- add preference to sendprivate tx request
- update tests

### DIFF
--- a/server/request_handler.go
+++ b/server/request_handler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/ethereum/go-ethereum/log"
@@ -58,11 +59,13 @@ func (r *RpcRequestHandler) process() {
 		(*r.respw).WriteHeader(http.StatusUnauthorized)
 		return
 	}
+
 	var isFast bool
-	var preference *types.Preference
-	if r.req.URL.Path == "/fast" {
+	var preferences *types.Preferences
+	if strings.Trim(r.req.URL.Path, "/") == "fast" { // If fast called, do not include tx to bundle, directly send tx to miners
 		isFast = true
-		preference = &types.Preference{Fast: &isFast}
+		preferences = &types.Preferences{Fast: &isFast}
+		r.logger.Info("[process] Setting fast preference", "isFast", isFast)
 	}
 
 	// If users specify a proxy url in their rpc endpoint they can have their requests proxied to that endpoint instead of Infura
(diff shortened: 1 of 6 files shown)


Ruteri (approved):

server/request_handler.go
```diff
@@ -59,6 +60,14 @@ func (r *RpcRequestHandler) process() {
 		return
 	}
 
+	var isFast bool
```
Ruteri, line 63: Nit: I'm guessing this can be declared inside if block

bhakiyakalimuthu: Done.

types/types.go
```diff
@@ -103,3 +104,13 @@ type BundleResponse struct {
 	BundleId string   `json:"bundleId"`
 	RawTxs   []string `json:"rawTxs"`
 }
+
+type SendPrivateTransactionRequestWithPreference struct {
+	SendPrivateTx flashbotsrpc.FlashbotsSendPrivateTransactionRequest
```
metachris, line 109:
```suggestion
	flashbotsrpc.FlashbotsSendPrivateTransactionRequest
```

types/types.go
```diff
@@ -103,3 +104,13 @@ type BundleResponse struct {
 	BundleId string   `json:"bundleId"`
 	RawTxs   []string `json:"rawTxs"`
 }
+
+type SendPrivateTransactionRequestWithPreference struct {
```
metachris, line 108:
```suggestion
type SendPrivateTransactionRequestWithPreferences struct {
```

types/types.go
```diff
@@ -103,3 +104,13 @@ type BundleResponse struct {
...
+	SendPrivateTx flashbotsrpc.FlashbotsSendPrivateTransactionRequest
+	Preferences   *Preferences `json:"preference,omitempty"`
+}
+
+type Preferences struct {
```
metachris, line 113:
```suggestion
type PrivateTxPreferences struct {
```

testutils/mock_rpcbackend.go
```diff
@@ -67,7 +68,11 @@ func handleRpcRequest(req *types.JsonRpcRequest) (result interface{}, err error)
 
 		// Relay calls
 	case "eth_sendPrivateTransaction":
-		param := req.Params[0].(map[string]interface{})
+		reqParam := req.Params[0].(map[string]interface{})
+		param := reqParam["SendPrivateTx"].(map[string]interface{})
+		if preference, ok := reqParam["preference"]; ok {
```
metachris, line 73: use plural

server/request_handler.go
metachris: it's not needed at all

server/request_handler.go
```diff
@@ -59,6 +60,13 @@ func (r *RpcRequestHandler) process() {
 		return
 	}
 
+	var preferences *types.Preferences
```
metachris, line 63: wouldn't even need to be a pointer, could instantiate here already and have every private tx have preferences. wdyt? 🤔

server/request_handler.go
```diff
@@ -59,6 +60,13 @@ func (r *RpcRequestHandler) process() {
...
+	if strings.Trim(r.req.URL.Path, "/") == "fast" { // If fast called, do not include tx to bundle, directly send tx to miners
```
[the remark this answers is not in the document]
bhakiyakalimuthu: I want to make sure if the user uses double slashes `baseUrl/fast/`, we should be able to handle

server/request_handler.go
```diff
@@ -59,6 +60,13 @@ func (r *RpcRequestHandler) process() {
...
+		isFast := true
+		preferences = &types.Preferences{Fast: &isFast}
```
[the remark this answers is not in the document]
bhakiyakalimuthu: I want to keep it as pointer as going forward we will have multiple preferences.

server/request_handler.go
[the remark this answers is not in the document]
metachris: makes sense!

types/types.go
```diff
@@ -103,3 +104,13 @@ type BundleResponse struct {
...
+	// If set do not include tx to bundle, directly send tx to miners
+	Fast *bool `json:"fast,omitempty"`
```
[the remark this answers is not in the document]
bhakiyakalimuthu: I want to keep it as pointer as going forward we will have multiple preferences. When we have auction included we want to omit fast.What do you think ?

server/request_handler.go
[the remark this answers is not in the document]
metachris: why does a pointer help? seems to me like an argument for not being a pointer and already instantiated 🤔

types/types.go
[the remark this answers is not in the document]
metachris: I think it should be either true or false and always sent. What's an advantage of having this be a pointer?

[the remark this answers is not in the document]
bhakiyakalimuthu: ok.makes sense

server/request_handler.go
bhakiyakalimuthu: updated

metachris (approved):
lgtm

PR #72, opened 2022-03-18
Block tx if nonce too low

metachris:


bhakiyakalimuthu (approved):

.github/workflows/test.yml
```diff
@@ -1,23 +1,42 @@
...
+  lint:
+    name: Lint
     runs-on: ubuntu-latest
     steps:
-    - name: Install Go
-      uses: actions/setup-go@v2
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.17
```
bhakiyakalimuthu, line 29: Is it supposed to be 1.18 right ?

metachris: yep, good catch, fixed :)

PR #73, opened 2022-03-20
[WIP] add batch db insert

bhakiyakalimuthu:
- Batch request db insert

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -9,6 +9,7 @@ import (
 	"github.com/flashbots/rpc-endpoint/server"
 	"os"
 	"strings"
+	"time"
 )
 
 var (
@@ -29,9 +30,14 @@ var (
 	redisUrl        = flag.String("redis", getEnvOrDefault("REDIS_URL", defaultRedisUrl), "URL for Redis (use 'dev' to use integrated in-memory redis)")
 	relayUrl        = flag.String("relayUrl", getEnvOrDefault("RELAY_URL", defaultRelayUrl), "URL for relay")
 	relaySigningKey = flag.String("signingKey", os.Getenv("RELAY_SIGNING_KEY"), "Signing key for relay requests")
-	psqlDsn         = flag.String("psql", os.Getenv("POSTGRES_DSN"), "Postgres DSN")
+	psqlDsn         = flag.String("psql", os.Getenv("POSTGRES_DSN_LOCAL"), "Postgres DSN")
 	debugPtr        = flag.Bool("debug", defaultDebug, "print debug output")
 	logJSONPtr      = flag.Bool("log-json", defaultLogJSON, "log in JSON")
+)
+
+const (
+	maxItem    = 100
+	maxTimeOut = time.Second * 5
 )
 
 func main() {
@@ -82,11 +88,13 @@ func main() {
 		db = database.NewPostgresStore(*psqlDsn)
 	}
 
+	pusher := server.NewRequestPusher(db, maxItem, maxTimeOut)
 	// Start the endpoint
-	s, err := server.NewRpcEndPointServer(version, *listenAddress, *proxyUrl, *relayUrl, key, *redisUrl, db)
+	s, err := server.NewRpcEndPointServer(version, *listenAddress, *proxyUrl, *relayUrl, key, *redisUrl, pusher)
 	if err != nil {
 		log.Crit("Server init error", "error", err)
 	}
+	go pusher.Run()
 	s.Start()
 }
 
--- a/database/mem_store.go
+++ b/database/mem_store.go
@@ -1,36 +1,34 @@
 package database
 
 import (
-	"github.com/google/uuid"
 	"sync"
 )
 
 type memStore struct {
-	Requests      map[uuid.UUID]RequestEntry
-	EthSendRawTxs map[uuid.UUID][]*EthSendRawTxEntry
+	Requests      []RequestEntry
+	EthSendRawTxs []*EthSendRawTxEntry
 	mutex         sync.Mutex
 }
 
 func NewMemStore() *memStore {
 	return &memStore{
-		Requests:      make(map[uuid.UUID]RequestEntry),
-		EthSendRawTxs: make(map[uuid.UUID][]*EthSendRawTxEntry),
-		mutex:         sync.Mutex{},
+		mutex: sync.Mutex{},
 	}
 }
 
-func (m *memStore) SaveRequestEntry(entry RequestEntry) error {
+func (m *memStore) SaveRequestEntries(entries []RequestEntry) error {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
-	m.Requests[entry.Id] = entry
+	m.Requests = append(m.Requests, entries...)
 	return nil
+
 }
 
-func (m *memStore) SaveRawTxEntries(entries []*EthSendRawTxEntry) error {
-	if len(entries) != 0 {
-		m.mutex.Lock()
-		defer m.mutex.Unlock()
-		m.EthSendRawTxs[entries[0].RequestId] = entries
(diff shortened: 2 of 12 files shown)


database/postgres_store.go
```diff
@@ -30,19 +34,81 @@ func (d *postgresStore) Close() {
...
+func (d *postgresStore) SaveRequestEntries(entries []RequestEntry) error {
+	tx, err := d.DB.Beginx()
+	var (
+		valStrings []string
+		valArgs    []interface{}
+	)
+	count := reflect.TypeOf(RequestEntry{}).NumField()
+	for i, entry := range entries {
+		valStrings = append(valStrings, fmt.Sprintf("($%d,$%d,$%d,$%d,$%d,$%d,$%d,$%d,$%d,$%d,$%d,$%d,$%d,$%d)",
+			i*count+1, i*count+2, i*count+3, i*count+4, i*count+5, i*count+6, i*count+7, i*count+8, i*count+9, i*count+10, i*count+11, i*count+12, i*count+13, i*count+14))
```
bhakiyakalimuthu, line 46: Not finding right way to handle this.

bhakiyakalimuthu:
Test needs to be fixed

database/postgres_store.go
Ruteri: You can just call `NamedExecContext` with a slice of structs
```
query = "insert into rpc_endpoint_requests (id, received_at, inserted_at, request_duration_ms, ...) values (:id, :received_at, ...)
d.DB.NamedExecContext(ctx, query, entries)
```

Doesn't it work as expected?

server/request_pusher.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (r *RequestPusher) Run() {
+	go r.saveEntries()
+	go r.addEntryToEntryQueue()
+}
+
+func (r *RequestPusher) addEntryToEntryQueue() {
+	for keepGoing := true; keepGoing; {
+		var batch []database.Entry
+		ticker := time.NewTicker(r.maxTimeOut)
```
Ruteri, line 36: We've seen issues when creating timers in a loop in the past - they are not cleaned up properly or fast enough and cause resource bleed resulting in scheduler running too slow (~80% cpu spent in the scheduler iterating old timers).
You could declare the ticker outside and reset it instead of creating a new one each iteration.

server/request_pusher.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+		}
+	push:
+		if len(batch) > 0 {
+			r.BatchChan <- batch
+		}
+	}
+}
+func (r *RequestPusher) saveEntries() {
+	for entry := range r.BatchChan {
```
Ruteri, line 59: I'm not sure if limiting to a single inserting thread won't result in degraded performance - the bigger the batch the longer you spend waiting for it and there's only so much items you can process at once.
Would just spinning up 4 goroutines doing this work? If so let's keep it in mind.

cmd/server/main.go
```diff
@@ -29,11 +30,16 @@ var (
 	redisUrl        = flag.String("redis", getEnvOrDefault("REDIS_URL", defaultRedisUrl), "URL for Redis (use 'dev' to use integrated in-memory redis)")
 	relayUrl        = flag.String("relayUrl", getEnvOrDefault("RELAY_URL", defaultRelayUrl), "URL for relay")
 	relaySigningKey = flag.String("signingKey", os.Getenv("RELAY_SIGNING_KEY"), "Signing key for relay requests")
-	psqlDsn         = flag.String("psql", os.Getenv("POSTGRES_DSN"), "Postgres DSN")
+	psqlDsn         = flag.String("psql", os.Getenv("POSTGRES_DSN_LOCAL"), "Postgres DSN")
 	debugPtr        = flag.Bool("debug", defaultDebug, "print debug output")
```
metachris, line 34: Send like a testing leftover

cmd/server/main.go
```diff
@@ -29,11 +30,16 @@ var (
...
 	logJSONPtr      = flag.Bool("log-json", defaultLogJSON, "log in JSON")
 )
 
+const (
+	maxItem    = 100
```
metachris, line 39: the names should be better. maxItem of what? perhaps those should just be defined inside the pusher for now.

cmd/server/main.go
```diff
@@ -82,11 +88,13 @@ func main() {
 		db = database.NewPostgresStore(*psqlDsn)
 	}
 
+	pusher := server.NewRequestPusher(db, maxItem, maxTimeOut)
```
metachris, line 91: why can't this be `db.StartRequestPusher()`?

database/mem_store.go
```diff
@@ -1,36 +1,34 @@
...
 	return nil
+
 }
 
-func (m *memStore) SaveRawTxEntries(entries []*EthSendRawTxEntry) error {
-	if len(entries) != 0 {
-		m.mutex.Lock()
-		defer m.mutex.Unlock()
-		m.EthSendRawTxs[entries[0].RequestId] = entries
+func (m *memStore) SaveRawTxEntries(entries [][]*EthSendRawTxEntry) error {
```
metachris, line 27: can you please add a docstring? why is entries a slice inside a slice?

server/request_pusher.go
bhakiyakalimuthu: ok.makes sense.

database/postgres_store.go
```diff
@@ -30,19 +34,81 @@ func (d *postgresStore) Close() {
 	d.DB.Close()
 }
 
-func (d *postgresStore) SaveRequestEntry(entry RequestEntry) error {
-	query := `INSERT INTO rpc_endpoint_requests
-	(id, received_at, request_duration_ms, is_batch_request, num_request_in_batch, http_method, http_url, http_query_param, http_response_status, ip_hash, origin, host, error) VALUES (:id, :received_at, :request_duration_ms, :is_batch_request, :num_request_in_batch, :http_method, :http_url, :http_query_param, :http_response_status, :ip_hash, :origin, :host, :error)`
+func (d *postgresStore) SaveRequestEntries(entries []RequestEntry) error {
```
metachris, line 37: I feel this function does too much magic. Please add comments and a function docstring.

server/request_pusher.go
```diff
@@ -0,0 +1,81 @@
...
+		maxTimeOut: maxTimeOut,
+	}
+}
+
+func (r *RequestPusher) Run() {
+	go r.saveEntries()
+	go r.addEntryToEntryQueue()
+}
+
+func (r *RequestPusher) addEntryToEntryQueue() {
```
metachris, line 33: This function is a bit of a mess :)

Usually it's best to avoid goto, except you need it. creating a new ticker every loop round is not needed (maybe a timer, but best to wait until the job is done before creating the next one).

also `keepGoing` is unnecessary, can just be a `for` loop and you break or return.

database/types.go
```diff
@@ -46,3 +46,8 @@ type EthSendRawTxEntry struct {
 	TxData                      string    `db:"tx_data"`
 	TxSmartContractMethod       string    `db:"tx_smart_contract_method"`
 }
+
+type Entry struct {
```
metachris, line 50: `Entry` seems too generic of a name. maybe `EntryContainer` or `RequestEntryContainer` describes a bit better what it does

cmd/server/main.go
bhakiyakalimuthu: Yes I did. But for testing,Its good to pass it while constructing.Also I have added more comments.

cmd/server/main.go
bhakiyakalimuthu: can you clarify here ? Because I want to keep business and persistant layer. Don't want add batch logic inside db though

server/request_pusher.go
bhakiyakalimuthu: I thought about using workerpool up front. then decided to keep one to see the performance.

metachris:
closing this until we revisit at a later date

PR #88, opened 2022-07-07
close response body

bhakiyakalimuthu:


server/util.go
```diff
@@ -116,6 +116,7 @@ func IsMetamaskMoz(r *http.Request) bool {
 }
 
 func ParseJsonRPCResponse(resp *http.Response) (*types.JsonRpcResponse, error) {
+	defer resp.Body.Close()
```
Ruteri, line 119: Can't it just be `resp.Body.Close()` after `ReadAll`?

Ruteri (commented):
The flow is what's causing the issue.
It's unclear who owns the `resp` object, and it's not clear who should `Close()` it.
I'm not entirely sure it should be `ParseJsonRPCResponse`, as it doesn't follow either the name or the responsibility or the ownership. I'd close the body one level up and just pass in bytes returned by `ioutil.ReadAll(resp.Body)`

PR #89, opened 2022-07-19
deprecate batch

bhakiyakalimuthu:
- deprecate handling batch requests
- disable go vet and static check for unreachable code

### DIFF
--- a/server/request_handler.go
+++ b/server/request_handler.go
@@ -40,6 +40,7 @@ func NewRpcRequestHandler(respw *http.ResponseWriter, req *http.Request, proxyUr
 	}
 }
 
+//nolint
 func (r *RpcRequestHandler) process() {
 	r.logger = log.New(log.Ctx{"uid": r.uid})
 	r.logger.Info("[process] POST request received")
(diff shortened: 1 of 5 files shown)


server/request_handler.go
```diff
@@ -101,6 +102,13 @@ func (r *RpcRequestHandler) process() {
 	// Parse JSON RPC payload
 	var jsonReq *types.JsonRpcRequest
 	if err = json.Unmarshal(body, &jsonReq); err != nil {
+		r.requestRecord.UpdateRequestEntry(r.req, http.StatusBadRequest, err.Error())
```
metachris, line 105: every single batch request will show up as error in the database. is this actually helpful?

server/request_handler.go
```diff
@@ -101,6 +102,13 @@ func (r *RpcRequestHandler) process() {
...
+		r.logger.Warn("[process] Parse payload", "error", err)
+		(*r.respw).WriteHeader(http.StatusBadRequest)
+		return
+
+		// Deprecate handling batch - DO NOT PROCEED
+		// Note: go vet checks are disabled for unreachable code
```
metachris, line 111: you could keep the linter and comment the code. but not important i think :)

server/request_handler.go
bhakiyakalimuthu: lets remove for now

server/request_handler.go
bhakiyakalimuthu: There are lot of code need to be commented.Linter does't allow. better to suppress linter warnings

metachris (approved):

PR #90, opened 2022-08-08
update ofac black list

bhakiyakalimuthu:


Ruteri (approved):

Ruteri:
in the relay we check for lowercase of the address, maybe it'd make sense to do it in rpc too

PR #102, opened 2022-09-16
docker log

bhakiyakalimuthu:
- replace iamge to alpine
- add more logging


Dockerfile
```diff
@@ -1,9 +1,11 @@
-FROM golang:1:18-alpine as builder
+# syntax=docker/dockerfile:1
+FROM golang:1.18-alpine as builder
 WORKDIR /build
 ADD . /build
-RUN --mount=type=cache,target=/root/.cache/go-build make build-for-docker
```
metachris, line 4: don't remove the `--mount` part as it's used for caching

PR #103, opened 2022-09-17
add service name as part of logging

bhakiyakalimuthu:


cmd/server/main.go
```diff
@@ -23,6 +24,7 @@ var (
 	defaultProxyTimeoutSeconds = 10
 	defaultRelayUrl            = "https://relay.flashbots.net"
 	defaultRedisUrl            = "localhost:6379"
+	defaultServiceName         = getEnvAsStrOrDefault("Service_Name", "rpc-endpoint")
```
metachris, line 27: env vars should always be capitalized
```suggestion
	defaultServiceName         = getEnvAsStrOrDefault("SERVICE_NAME", "rpc-endpoint")
```

cmd/server/main.go
```diff
@@ -35,6 +37,7 @@ var (
 	psqlDsn             = flag.String("psql", os.Getenv("POSTGRES_DSN"), "Postgres DSN")
 	debugPtr            = flag.Bool("debug", defaultDebug, "print debug output")
 	logJSONPtr          = flag.Bool("log-json", defaultLogJSON, "log in JSON")
+	ServiceName         = flag.String("Service-Name", defaultServiceName, " ServiceName/deployment name")
```
metachris, line 40: flag should be lowercase. also please improve the help text, it's not clear from that alone what this flag does

cmd/server/main.go
```diff
@@ -35,6 +37,7 @@ var (
 	psqlDsn             = flag.String("psql", os.Getenv("POSTGRES_DSN"), "Postgres DSN")
 	debugPtr            = flag.Bool("debug", defaultDebug, "print debug output")
 	logJSONPtr          = flag.Bool("log-json", defaultLogJSON, "log in JSON")
+	serviceName         = flag.String("serviceName", defaultServiceName, "name of the service which will be used in the logs")
```
metachris, line 40: ensure consistency
```suggestion
	serviceName         = flag.String("service-name", defaultServiceName, "name of the service which will be used in the logs")
```

cmd/server/main.go
```diff
@@ -53,28 +56,29 @@ func main() {
 	}
 
 	log.Root().SetHandler(log.LvlFilterHandler(logLevel, log.StreamHandler(os.Stderr, logFormat)))
+	logger := log.New(log.Ctx{"service": *serviceName})
```
metachris, line 59: maybe only do this if serviceName is not empty? perhaps some people don't want it, then could disable it with using an empty serviceName 🤔

bhakiyakalimuthu: we have already set the default value as rpc-endpoint. Not sure this will be interesting one for the users

metachris (approved):

PR #115, opened 2023-06-06
[fix] Implement pod draining

0x416e746f6e:
This PR implements separate HTTP server listening on `DRAIN_ADDR`.  When it receives HTTP GET, it will flag server as un-healthy.  From that moment on `/health` endpoint of the main HTTP server will start reporting `HTTP 500`.

If we use this in conjunction with k8s pre-stop hook, we can make sure that on updates/restarts the pod lives long enough to serve whatever requests the load-balancer would still forward to it until it detects that the server is not healthy.

For example:

```yaml
          lifecycle:
            preStop:
              httpGet:
                port: 8081
          env:
            - name: DRAIN_ADDR
              value: :8081
```

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -20,6 +20,7 @@ var (
 	defaultDebug               = os.Getenv("DEBUG") == "1"
 	defaultLogJSON             = os.Getenv("LOG_JSON") == "1"
 	defaultListenAddress       = "127.0.0.1:9000"
+	defaultDrainAddress        = "127.0.0.1:9001"
 	defaultProxyUrl            = "http://127.0.0.1:8545"
 	defaultProxyTimeoutSeconds = 10
 	defaultRelayUrl            = "https://relay.flashbots.net"
@@ -29,6 +30,7 @@ var (
 	// cli flags
 	versionPtr          = flag.Bool("version", false, "just print the program version")
 	listenAddress       = flag.String("listen", getEnvAsStrOrDefault("LISTEN_ADDR", defaultListenAddress), "Listen address")
+	drainAddress        = flag.String("drain", getEnvAsStrOrDefault("DRAIN_ADDR", defaultDrainAddress), "Drain address")
 	proxyUrl            = flag.String("proxy", getEnvAsStrOrDefault("PROXY_URL", defaultProxyUrl), "URL for default JSON-RPC proxy target (eth node, Infura, etc.)")
 	proxyTimeoutSeconds = flag.Int("proxyTimeoutSeconds", getEnvAsIntOrDefault("PROXY_TIMEOUT_SECONDS", defaultProxyTimeoutSeconds), "proxy client timeout in seconds")
 	redisUrl            = flag.String("redis", getEnvAsStrOrDefault("REDIS_URL", defaultRedisUrl), "URL for Redis (use 'dev' to use integrated in-memory redis)")
@@ -92,7 +94,18 @@ func main() {
 		db = database.NewPostgresStore(*psqlDsn)
 	}
 	// Start the endpoint
-	s, err := server.NewRpcEndPointServer(logger, version, *listenAddress, *relayUrl, *proxyUrl, *proxyTimeoutSeconds, key, *redisUrl, db)
+	s, err := server.NewRpcEndPointServer(server.Configuration{
+		DB:                  db,
+		DrainAddress:        *drainAddress,
+		ListenAddress:       *listenAddress,
+		Logger:              logger,
+		ProxyTimeoutSeconds: *proxyTimeoutSeconds,
+		ProxyUrl:            *proxyUrl,
+		RedisUrl:            *redisUrl,
+		RelaySigningKey:     key,
+		RelayUrl:            *relayUrl,
+		Version:             version,
+	})
 	if err != nil {
 		logger.Crit("Server init error", "error", err)
 	}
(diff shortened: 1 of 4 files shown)


metachris:
Interesting! This PR does a lot and I don't have the resources to review the nuances of these changes right now. Will defer to @dvush and @Ruteri

server/server.go
```diff
@@ -64,19 +75,32 @@ func NewRpcEndPointServer(logger log.Logger, version, listenAddress, relayUrl, p
 	FlashbotsRPC = flashbotsrpc.New(relayUrl)
 	// FlashbotsRPC.Debug = true
 
-	return &RpcEndPointServer{
+	mux := http.NewServeMux()
```
0x416e746f6e, line 78: **explanation:**

move the routing setup into the `NewRpcEndPointServer` method.  this will make it possible to run more than one server (on different ports, for example).  also, it's nice to encapsulate the creation logic into one place.

0x416e746f6e:
oh, there's also https://github.com/flashbots/rpc-endpoint/pull/114 that somewhat overlaps.  sorry @dvush, I didn't see it before.  please let me know if I should refactor this PR to build on top of yours.

dvush:
@0x416e746f6e , yes adapt please. It already handles graceful shutdown except health endpoint part

0x416e746f6e:
> yes adapt please. It already handles graceful shutdown except health endpoint part

@dvush:  updated!  please have a look.

cmd/server/main.go
```diff
@@ -92,7 +94,18 @@ func main() {
 		db = database.NewPostgresStore(*psqlDsn)
 	}
 	// Start the endpoint
-	s, err := server.NewRpcEndPointServer(logger, version, *listenAddress, *relayUrl, *proxyUrl, *proxyTimeoutSeconds, key, *redisUrl, db)
+	s, err := server.NewRpcEndPointServer(server.Configuration{
```
0x416e746f6e, line 97: **explanation:**

instead of passing a train of arguments to the `NewRpcEndPointServer` (which would render it error-prone due to arguments being misplaced) let's use `struct` that will have all of them properly named.

server/server.go
```diff
@@ -142,6 +193,14 @@ func (s *RpcEndPointServer) HandleHttpRequest(respw http.ResponseWriter, req *ht
 	request.process()
 }
 
+func (s *RpcEndPointServer) handleDrain(respw http.ResponseWriter, req *http.Request) {
```
dvush, line 196: Who will call this? Pre stop hook?

I thought that we only need to handle SIGTERM and then start returning errors on  /health

server/server.go
```diff
@@ -142,6 +193,14 @@ func (s *RpcEndPointServer) HandleHttpRequest(respw http.ResponseWriter, req *ht
...
+	if s.isHealthy {
+		s.isHealthy = false
+		s.logger.Info("Server marked as unhealthy")
+		time.Sleep(60 * time.Second) // Give LB enough time to detect us unhealthy
```
dvush, line 200: So the idea is that request to the drain endpoint should take 60 seconds?

server/server.go
```diff
@@ -157,7 +216,11 @@ func (s *RpcEndPointServer) handleHealthRequest(respw http.ResponseWriter, req *
 	}
 
 	respw.Header().Set("Content-Type", "application/json")
-	respw.WriteHeader(http.StatusOK)
+	if s.isHealthy {
```
dvush, line 219: don't we need some kind of mutex/atomic for this?

server/server.go
```diff
@@ -90,35 +97,79 @@ func (s *RpcEndPointServer) Start() {
...
-	http.HandleFunc("/", s.HandleHttpRequest)
-	http.HandleFunc("/health", s.handleHealthRequest)
-	http.HandleFunc("/bundle", s.HandleBundleRequest)
+	s.startMainServer()
+	s.startDrainServer()
+
+	notifier := make(chan os.Signal, 1)
+	signal.Notify(notifier, os.Interrupt, syscall.SIGTERM)
+
+	<-notifier
```
dvush, line 106: My understanding is that we don't need to have separate server for that and call "drain" method.

All we need is to: 
1. start returning errors from /health after this line is crossed
2. Wait n-seconds here instead of in the drain handler

If its the same then I would prefer to do it here with one server because its more simple and straightforward. 

But maybe its not the same?

0x416e746f6e: I also thought about that idea (sleep in the `SIGTERM` handler).

reason I did not go for it was that we only need to do that (sleeping) when we run it in k8s behind AWS LB target group (because it takes its time to become aware of the pod being drained).  while in the case of plain k8s service, the service becomes aware of the pod being drained basically immediately (I guess the control plane first updates the service's endpoints, and only after sends a signal to the pod).

if we introduce the sleep into SIGTERM handler then we'd impose that on all workflows regardless if that would be necessary or not.

server/server.go
0x416e746f6e: yes.

reason is that control plane considers the pre-hook to be "done" right at the moment when the request returns a status, which is when it will send the SIGTERM to the pod.

server/server.go
0x416e746f6e: in the purest sense, maybe "yes".  but practically, not really.  we don't mind if there's race-condition here (if not this health-check sees the new status, then the next one will).  and it's ok if there is concurrent access (we don't play with pointers here, and we read the value of isHealthy only once in this method).

server/server.go
dvush: I mean something like this https://github.com/flashbots/rpc-endpoint/pull/116

server/server.go
0x416e746f6e: but maybe worthwhile to put this into a comment inside the source-code.  who knows what other functionality will be added here in the future, so the future maintainer should better be aware.

server/server.go
dvush: You probably right, I agree. Did not see your comment

dvush (approved):

PR #116, opened 2023-06-09
Improve graceful shutdown

dvush:
Inspired by https://github.com/flashbots/rpc-endpoint/pull/115

also handles second sigterm correctly

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -10,6 +10,7 @@ import (
 	"os/signal"
 	"runtime"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -41,6 +42,8 @@ type RpcEndPointServer struct {
 	proxyTimeoutSeconds int
 	relaySigningKey     *ecdsa.PrivateKey
 	db                  database.Store
+	healthy             bool
+	healthyMu           sync.Mutex
 }
 
 func NewRpcEndPointServer(logger log.Logger, version, listenAddress, relayUrl, proxyUrl string, proxyTimeoutSeconds int, relaySigningKey *ecdsa.PrivateKey, redisUrl string, db database.Store) (*RpcEndPointServer, error) {
@@ -76,6 +79,8 @@ func NewRpcEndPointServer(logger log.Logger, version, listenAddress, relayUrl, p
 		proxyTimeoutSeconds: proxyTimeoutSeconds,
 		relaySigningKey:     relaySigningKey,
 		db:                  db,
+		healthy:             true,
+		healthyMu:           sync.Mutex{},
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -112,9 +117,26 @@ func (s *RpcEndPointServer) Start() {
 
 	<-notifier
 
+	s.SetUnhealthy()
+	s.logger.Info("http server stopping")
+
+	// Wait for 60 seconds for load balancer to pick up the unhealthy state or for SIGTERM to be sent again
+	select {
+	case <-time.After(60 * time.Second):
+	case <-notifier:
```
0x416e746f6e, line 126: **question:**

why would the `SIGTERM` sent for a second time?  I think if the OS/k8s gets impatient (that is, when its graceful shutdown timeout expires) it would then just send `SIGKILL` and that would be the end of story.

maybe convenient for the developer though..  (they can then hit Ctrl-C twice and avoid waiting for 60s when they want to terminate the server running locally).

server/server.go
```diff
@@ -205,3 +234,9 @@ func setCorsHeaders(respw http.ResponseWriter) {
 	respw.Header().Set("Access-Control-Allow-Origin", "*")
 	respw.Header().Set("Access-Control-Allow-Headers", "Accept,Content-Type")
 }
+
+func (s *RpcEndPointServer) SetUnhealthy() {
+	s.healthyMu.Lock()
```
0x416e746f6e, line 239: **question (non-blocking):**

do we really need a whole mutex to protect something that is going to be changed only once?  (i.e. there will be only once that the transition from `true` to `false` will happen, no matter how many times `SetUnhealthy()` is called).

server/server.go
```diff
@@ -112,9 +117,26 @@ func (s *RpcEndPointServer) Start() {
 
 	<-notifier
 
+	s.SetUnhealthy()
+	s.logger.Info("http server stopping")
+
+	// Wait for 60 seconds for load balancer to pick up the unhealthy state or for SIGTERM to be sent again
+	select {
+	case <-time.After(60 * time.Second):
```
0x416e746f6e, line 125: **issue:**

now, we will have to wait for 60s every time we shut it down, even if it's run locally.

dvush: agree, fixed

server/server.go
dvush: I see what you mean, but I would better use mutex/atomic when you have separate threads every time. Because I don't know how compiler handles this. I vaguely remember that compilers can do something awful when handling situations like that. Especially bools.

server/server.go
dvush: Yes, its for Ctrl+C. Not a big deal but for completeness.

server/server.go
0x416e746f6e: unless AWS LB target group "talks" directly to the EC2's port.

but then, that would render the implementation AWS-specific.  e.g. I am not sure how exactly GCP with its VPC-native clusters handles the ingress.

server/server.go
0x416e746f6e: sure.  no way mutex could hurt here.

although, since `/health` is exposed to the internet, that might open a funny way to DDOS us via mutex contention.

server/server.go
0x416e746f6e: o.t.o.h., using pre-stop hook is the way AWS suggests to do it themselves:
https://aws.amazon.com/blogs/containers/how-to-rapidly-scale-your-application-with-alb-on-eks-without-losing-traffic/

dvush: From the example in this link I get a different conclusion.

Are you sure that "TERMINATING" state is relevant here? 

What I see there is that they use pre stop hook to trigger health unavailable response. But this health unavailable will lead to
>its corresponding k8s service stops routing any more requests to it

independent on if node is terminating or not because they use readiness probe.

Do you want to test it? We have both of prs ready so it will not take a long time.
We can do it together. I think it would be interesting. I would learn something

PR #118, opened 2023-06-20
Config refunds

dvush:
## 📝 Summary

Allow users to specify refund address / refund percentage for the mev share. 

## 📚 References

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/server/url_params.go
+++ b/server/url_params.go
@@ -13,11 +13,14 @@ import (
 var (
 	DefaultAuctionHint = []string{"hash"}
 
-	ErrEmptyHintQuery          = errors.New("Hint query must be non-empty if set.")
-	ErrEmptyTargetBuilderQuery = errors.New("Target builder query must be non-empty if set.")
-	ErrIncorrectAuctionHints   = errors.New("Incorrect auction hint, must be one of: contract_address, function_selector, logs, calldata.")
-	ErrIncorrectOriginId       = errors.New("Incorrect origin id, must be less then 255 char.")
-	ErrIncorrectRefundQuery    = errors.New("Incorrect refund.")
+	ErrEmptyHintQuery                      = errors.New("Hint query must be non-empty if set.")
+	ErrEmptyTargetBuilderQuery             = errors.New("Target builder query must be non-empty if set.")
+	ErrIncorrectAuctionHints               = errors.New("Incorrect auction hint, must be one of: contract_address, function_selector, logs, calldata.")
+	ErrIncorrectOriginId                   = errors.New("Incorrect origin id, must be less then 255 char.")
+	ErrIncorrectRefundQuery                = errors.New("Incorrect refund query, must be 0xaddress:percentage.")
+	ErrIncorrectRefundAddressQuery         = errors.New("Incorrect refund address.")
+	ErrIncorrectRefundPercentageQuery      = errors.New("Incorrect refund percentage.")
+	ErrIncorrectRefundTotalPercentageQuery = errors.New("Incorrect refund total percentage, must be bellow 100%.")
 )
 
 type URLParameters struct {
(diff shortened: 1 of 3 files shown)


server/url_params.go
```diff
@@ -63,6 +67,77 @@ func ExtractParametersFromUrl(url *url.URL) (params URLParameters, err error) {
 		params.pref.Builders = targetBuildersQuery
 	}
 
+	refundAddressQuery, ok := url.Query()["refund"]
+	if ok {
```
Wazzymandias, line 71: why not invert this and reduce the amount of nested code? i.e.

```
if !ok {
    return params, nil 
}

# rest of logic here 
```

server/url_params.go
```diff
@@ -63,6 +67,77 @@ func ExtractParametersFromUrl(url *url.URL) (params URLParameters, err error) {
...
+		if len(refundAddressQuery) == 0 {
+			return params, ErrIncorrectRefundQuery
```
Wazzymandias, line 73: **nitpick (non-blocking):** It would be nice if we had more granular error message, e.g.

```suggestion
			return params, ErrEmptyRefundAddressQuery
```

server/url_params.go
```diff
@@ -63,6 +67,77 @@ func ExtractParametersFromUrl(url *url.URL) (params URLParameters, err error) {
...
+		}
+
+		var (
+			addresses = make([]common.Address, len(refundAddressQuery))
+			percents  = make([]int, len(refundAddressQuery))
+		)
+
+		for i, refundAddress := range refundAddressQuery {
```
Wazzymandias, line 81: **suggestion:**  Can you add small, concise comment that provides **example for structure of refund address query**? 

A link to any existing spec would also be great. Right now it's not clear what we're expecting the valid input to look like.

server/url_params.go
```diff
@@ -63,6 +67,77 @@ func ExtractParametersFromUrl(url *url.URL) (params URLParameters, err error) {
...
+			split := strings.Split(refundAddress, ":")
+			if len(split) != 2 {
+				return params, ErrIncorrectRefundQuery
+			}
+			if !common.IsHexAddress(split[0]) {
+				return params, ErrIncorrectRefundQuery
```
Wazzymandias, line 87: **todo:**
```suggestion
				return params, ErrInvalidRefundAddress
```

server/url_params.go
```diff
@@ -63,6 +67,77 @@ func ExtractParametersFromUrl(url *url.URL) (params URLParameters, err error) {
...
+			}
+			address := common.HexToAddress(split[0])
+			percent, err := strconv.Atoi(split[1])
+			if err != nil {
+				return params, ErrIncorrectRefundQuery
```
Wazzymandias, line 92: **todo:**

```suggestion
				return params, ErrInvalidRefundPercentage
```

server/url_params.go
```diff
@@ -63,6 +67,77 @@ func ExtractParametersFromUrl(url *url.URL) (params URLParameters, err error) {
...
+		if totalRefund <= 0 || totalRefund >= 100 {
+			return params, ErrIncorrectRefundQuery
+		}
+
+		// normalize refund config percentages
+		for i, percent := range percents {
+			percents[i] = (percent * 100) / totalRefund
+		}
+
+		// should sum to 100
```
Wazzymandias, line 114: **praise:** I like the clear comments, helps make the code more understandable!

server/url_params_test.go
```diff
@@ -68,11 +73,57 @@ func TestExtractAuctionPreferenceFromUrl(t *testing.T) {
 			want: URLParameters{
 				pref:       types.TxPrivacyPreferences{Hints: []string{"hash", "special_logs"}, Builders: []string{"builder1", "builder2"}},
 				prefWasSet: false,
-				originId: "" +
-					"",
+				originId:   "",
 			},
 			err: nil,
 		},
+		"set refund": {
```
Wazzymandias, line 80: **praise:** great tests!

server/url_params.go
dvush: I think it is better like this because if you look above - its how other queries are done. This `!ok return` can work only for the last query. So if we add another query here we would still need to change to 'if ok` like here.

Wazzymandias (approved):

PR #119, opened 2023-06-28
set origin as a header

dvush:
## 📝 Summary

Adds support for "x-flashbots-origin" header on the relay.

## ⛱ Motivation and Context

x-flashbots-origin is a new header on the relay that lets user to specify order flow origin

## 📚 References

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/server/request_processor.go
+++ b/server/request_processor.go
@@ -31,6 +31,7 @@ type RpcRequest struct {
 	tx                         *ethtypes.Transaction
 	txFrom                     string
 	relaySigningKey            *ecdsa.PrivateKey
+	relayUrl                   string
 	origin                     string
 	referer                    string
 	isWhitehatBundleCollection bool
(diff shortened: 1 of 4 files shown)


server/request_processor.go
```diff
@@ -330,7 +337,9 @@ func (r *RpcRequest) handleCancelTx() (requestCompleted bool) {
 	}
 
 	cancelPrivTxArgs := flashbotsrpc.FlashbotsCancelPrivateTransactionRequest{TxHash: initialTxHash}
-	_, err = FlashbotsRPC.FlashbotsCancelPrivateTransaction(r.relaySigningKey, cancelPrivTxArgs)
+
+	fbRpc := flashbotsrpc.New(r.relayUrl)
```
metachris, line 341: does recreating the flashbotsrpc instance have negative performance impact, like not reusing the HTTP client connection pool?

metachris (approved):
lgtm. 

not sure if the flashbotsrpc instance should be reused for HTTP client connection pooling and reuse, but should be fine like this too.

Wazzymandias (commented):
@dvush a note: The use of `X-` as prefix to headers for non-standardized parameters has been discouraged since 2012 in [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648).

I won't block the PR for this but it would be nice if we move towards best practice in the future, especially while the number of custom headers we support is low. Otherwise if large number of existing application headers use `X-` it would be better to continue to do so for consistency.

PR #130, opened 2023-10-31
fix redirect link on GET endpoint

TymKh:
## 📝 Summary

Simple fix for outdated doc link

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -191,7 +191,7 @@ func (s *RpcEndPointServer) HandleHttpRequest(respw http.ResponseWriter, req *ht
 
 	if req.Method == http.MethodGet {
 		if strings.Trim(req.URL.Path, "/") == "fast" {
-			http.Redirect(respw, req, "https://docs.flashbots.net/flashbots-protect/rpc/fast-mode/", http.StatusFound)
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -191,7 +191,7 @@ func (s *RpcEndPointServer) HandleHttpRequest(respw http.ResponseWriter, req *ht
 
 	if req.Method == http.MethodGet {
 		if strings.Trim(req.URL.Path, "/") == "fast" {
-			http.Redirect(respw, req, "https://docs.flashbots.net/flashbots-protect/rpc/fast-mode/", http.StatusFound)
+			http.Redirect(respw, req, "https://docs.flashbots.net/flashbots-protect/quick-start#faster-transactions", http.StatusFound)
```
0x416e746f6e, line 194: **question:**

should this really be hard-coded?  why not, e.g., use ENV var for this and configure the workload on infra side?

TymKh: It's definitely true what you say, but it's supposed to be once in a lifetime change. If it happens next time, I will refactor it to be configurable

PR #137, opened 2024-04-06
make build works on macs

samlaf:
Currently hardcoded to build for linux platforms. This change makes build also able to target mac computers by dynamically checking the uname


metachris:
thanks for the effort. how about removing GOOS completely instead of adding the logic to determine?

PR #141, opened 2024-05-23
make URL parameters case-insensitive

odysseus0:
## 📝 Summary

This PR makes RPC URL parameter case-insensitive.

## ⛱ Motivation and Context

We had had multiple instances where clients who claimed to have added `originId` parameters but failed to appear in our DB.

It turns out that many of them simply spelled it wrong, like `orignid` instead of `originId`.

This PR makes the RPC more robust to this type of error.

## 📚 References

None. Mostly observation of clients using `originId` with wrong cases.

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/server/url_params.go
+++ b/server/url_params.go
@@ -31,6 +31,17 @@ type URLParameters struct {
 	fast       bool
 }
 
+// normalizeQueryParams takes a URL and returns a map of query parameters with all keys normalized to lowercase.
+// This helps in making the parameter extraction case-insensitive.
+func normalizeQueryParams(url *url.URL) map[string][]string {
+	normalizedQuery := make(map[string][]string)
+	for key, values := range url.Query() {
+		normalizedKey := strings.ToLower(key)
+		normalizedQuery[normalizedKey] = values
+	}
+	return normalizedQuery
+}
+
 // ExtractParametersFromUrl extracts the auction preference from the url query
 // Allowed query params:
 //   - hint: mev share hints, can be set multiple times, default: hash, special_logs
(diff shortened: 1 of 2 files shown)


metachris (approved):
nice code, thanks!

ryanschneider (approved):
lgtm!

Divinefavo1 (approved):

server/url_params.go
```diff
@@ -11,17 +11,49 @@ import (
...
+		"calldata":          {},
+		"default_logs":      {},
+	}
+
+	ErrEmptyHintQuery          = errors.New("hint query must be non-empty if set")
+	ErrEmptyTargetBuilderQuery = errors.New("target builder query must be non-empty if set")
+	ErrIncorrectAuctionHints   = func(hint string) error {
+		return fmt.Errorf("incorrect auction hint: %s, must be one of: contract_address, function_selector, logs, calldata, default_logs", hint)
+	}
+	ErrIncorrectOriginId = func(originId string) error {
```
TymKh, line 46: This is non-idiomatic way of declaring errors in go due to the fact that you can't leverage `errors.Is` this way. There should be constant underlying `ErrIncorrectOriginID=fmt.Errof("incorrect origin id, must be non-empty and less than 255 characters")`. If you want additional context you use `fmt.Errorf("%w value: %s, ErrIncorrectOriginID, originID)`

TymKh:
Looks good, just posted a note regarding error management

Divinefavo1 (commented):
How do this seems to e
Work

metachris:
Puh, I think the changes to the errors are going a bit overboard. Is this really needed?

odysseus0:
Reverted and kept the PR to bare minimal.

For future reference, will keep PR focused and separate code changes like refactoring into their own PR.

Divinefavo1 (commented):
Alright

Divinefavo1 (commented):
I have received massage Please can you refund me

PR #142, opened 2024-05-29
Log request url for eth_sendRawTransaction

odysseus0:
## 📝 Summary

Log request url for eth_sendRawTransaction

## ⛱ Motivation and Context

Clients have been quite error prone when it comes to setting URL parameters. Currently there isn't one source of truth we can check as we don't log or record in DB the actual URL used by user. Having logging of the exact URL used would greatly facilitate debugging.

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/server/request_handler.go
+++ b/server/request_handler.go
@@ -134,7 +134,7 @@ func (r *RpcRequestHandler) processRequest(client RPCProxyClient, jsonReq *types
 	if jsonReq.Method == "eth_sendRawTransaction" {
 		entry = r.requestRecord.AddEthSendRawTxEntry(uuid.New())
 		// log the full url for debugging
-		r.logger.Info("[processRequest] Request URL", "url", reqURL)
+		r.logger.Info("[processRequest] eth_sendRawTransaction request URL", "url", reqURL)
 	}
 	// Handle single request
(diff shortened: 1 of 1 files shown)


server/request_handler.go
```diff
@@ -125,14 +125,16 @@ func (r *RpcRequestHandler) process() {
...
 }
 
 // processRequest handles single request
-func (r *RpcRequestHandler) processRequest(client RPCProxyClient, jsonReq *types.JsonRpcRequest, origin, referer string, isWhitehatBundleCollection bool, whitehatBundleId string, urlParams URLParameters) {
+func (r *RpcRequestHandler) processRequest(client RPCProxyClient, jsonReq *types.JsonRpcRequest, origin, referer string, isWhitehatBundleCollection bool, whitehatBundleId string, urlParams URLParameters, reqURL string) {
 	var entry *database.EthSendRawTxEntry
 	if jsonReq.Method == "eth_sendRawTransaction" {
 		entry = r.requestRecord.AddEthSendRawTxEntry(uuid.New())
+		// log the full url for debugging
+		r.logger.Info("[processRequest] Request URL", "url", reqURL)
```
metachris, line 137: why pass in reqURL, instead of logging this before line 128?

metachris, line 137:
```suggestion
		r.logger.Info("[processRequest] eth_sendRawTransaction request URL", "url", reqURL)
```

metachris (approved):

PR #148, opened 2024-07-10
feat(debug): Log a privacy preserving hash of IP and UserAgent to assist in rate limiting debugging

ryanschneider:
## 📝 Summary

We are seeing abuse of our endpoint that is bypassing our firewall rules.  To help determine how they are getting past the firewall we need to log the "source" of the traffic in a privacy-preserving way.

~What I came up with is the `xxhash( x-forwarded-for + user-agent )` or more specifically the left-most non-private IP in the `X-Forwarded-For` and the `User-Agent` headers.~

However, after discussing w/ @0x416e746f6e he made the good point that `User-Agent` can be gamed by randomly generating a new UA w/ each request.  Furthermore, the fingerprint should *not* be long-term stable, as that would allow a malicious actor with access to Flashbot logs the ability to track user behavior long-term.  So now the hash is salted w/ the current timestamp truncated to the hour, and the UA is removed.

In addition, we currently use [`proxyd`](https://github.com/ethereum-optimism/infra/tree/main/proxyd) as our upstream which by default uses the `X-Forwarded-For` header for rate limiting.  Since `xxhash` is a `uint64` it actually fits perfectly in an IPv6 address, so we convert the fingerprint to a fake IPv6 address in the reserved "example address" documentation prefix from https://datatracker.ietf.org/doc/html/rfc3849 and insert this "IP" as the `X-Forwarded-For` field.

## ⛱ Motivation and Context

Give us to means to perform log aggregation to see if the offending traffic is coming from a single or multiple sources.

## 📚 References

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/server/fingerprint.go
+++ b/server/fingerprint.go
@@ -39,7 +39,16 @@ func init() {
 // FingerprintFromRequest returns a fingerprint for the request based on the X-Forwarded-For header
 // and a salted timestamp.  The fingerprint is used to identify unique users sessions
 // over a short period of time, and thus can be used as a key for rate limiting.
-func FingerprintFromRequest(req *http.Request, at time.Time) (Fingerprint, error) {
+// The seed param is additional entropy to make the fingerprint resistant to rainbow table lookups.
+// Without the seed a malicious rpc operator could reverse a client IP address to a fingerprint
+// by exhausting all possible IP addresses and comparing the resulting fingerprints.
+//
+// We considered adding the User-Agent header to the fingerprint, but decided
+// against it because it would make the fingerprint gameable.  Instead, we
+// will salt the fingerprint with the current timestamp rounded to the
+// latest hour. This will make sure fingerprints rotate every hour so we
+// cannot reasonably track user behavior over time.
+func FingerprintFromRequest(req *http.Request, at time.Time, seed uint64) (Fingerprint, error) {
 	// X-Forwarded-For header contains a comma-separated list of IP addresses when
 	// the request has been forwarded through multiple proxies.  For example:
 	//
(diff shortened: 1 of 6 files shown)


server/fingerprint.go
```diff
@@ -0,0 +1,113 @@
...
+func FingerprintFromRequest(req *http.Request, at time.Time) (Fingerprint, error) {
+	// X-Forwarded-For header contains a comma-separated list of IP addresses when
+	// the request has been forwarded through multiple proxies.  For example:
+	//
+	// X-Forwarded-For: 2600:8802:4700:bee:d13c:c7fb:8e0f:84ff, 172.70.210.100
+	xff, err := getXForwardedForIP(req)
+	if err != nil {
+		return 0, err
+	}
+	// We considered adding the User-Agent header to the fingerprint, but decided
```
metachris, line 27: this could be part of the overall function comment as it's important and general

server/fingerprint.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func getXForwardedForIP(r *http.Request) (string, error) {
+	// gets the left-most non-private IP in the X-Forwarded-For header
+	xff := r.Header.Get("X-Forwarded-For")
+	if xff == "" {
+		return "", fmt.Errorf("no X-Forwarded-For header")
+	}
+	ips := strings.Split(xff, ",")
+	for _, ip := range ips {
```
metachris, line 68: do we also need to truncate the IP here? some might have a space in them if it's sent with a space

server/fingerprint.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	return false
+}
+
+// Taken from https://github.com/tomasen/realip/blob/master/realip.go
+// MIT Licensed, Copyright (c) 2018 SHEN SHENG
+var cidrs []*net.IPNet
+
+func init() {
+	maxCidrBlocks := []string{
```
metachris, line 97: this could be a variable up top

metachris (approved):
code looks great. left a few minor comment but nothing important

metachris (approved):

PR #149, opened 2024-08-01
chore(golang): Update to Go 1.22

ryanschneider:
## 📝 Summary

Update `go.mod` to go 1.21+ format w/ `toolchain` directive and update Docker container to use go 1.22.

## ⛱ Motivation and Context

Running `go mod tidy` w/ go 1.22 or higher wants to rewrite the `go.mod` with the new `toolchain` directive, which causes go 1.20 to die with an unrecognized directive error.

## 📚 References

https://go.dev/doc/toolchain

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`


ryanschneider:
Forgot to mention this also updates geth to 1.14.7 which is the minimum supported version for go 1.21 and higher.

go.mod
```diff
@@ -1,17 +1,20 @@
 module github.com/flashbots/rpc-endpoint
 
-go 1.20
+go 1.21
```
metachris, line 3: is it necessary to make the _minimum_ version v1.21?

metachris (approved):

go.mod
ryanschneider: Ya, versions older than 1.21 choke on the new `toolchain` directive below that go 1.22 adds.  I'm actually surprised that the Go team didn't consider this, I feel like the toolchain directive should've only been added for newer `go` versions but I can't find a way to do that.

Basically running into the same issue as: https://github.com/golang/go/issues/65847#issuecomment-1957608903

Issue #150, opened 2024-08-01
Proposal: return `eth_getTransactionCount` from the private mempool when RPC request is signed


ryanschneider:
Proposal: return `eth_getTransactionCount` from the private mempool when RPC request is signed
============================================================================

This is a proposal to augment the RPC endpoint behavior as follows:

- When a request is made for `eth_getTransactionCount` targeting the `"pending"` block specifier.
- And the request is signed with the `X-Flashbots-Signature` header.
- And the signature address matches the account being queried
- Then the RPC endpoint will return the transaction count from it's internal "private mempool".

### Current Behavior

Recall that `eth_getTransactionCount` takes two arguments:

- The account to query
- The EIP-1898 Block Specifier (e.g. `"latest"`, `"pending"` or a block number or hash).

And returns the total number of transactions that account has executed at that block specifier.  Furthermore, the `"pending"` specifier is somewhat special-cased, in that in addition to considering all transactions already executed on-chain, it also includes any pending transactions in the node's mempool that are considered _executable_ by the node, where _executable_ refers means that all previous nonces for the account in question are either already on-chain or themselves executable.

Currently, `eth_getTransactionCount` is treated like any other RPC not directly handled by the endpoint and the request is simply proxied to the `PROXY_URL` endpoint.  However, this endpoint is typically not aware of any "private mempool" transactions sent by the user (e.g. `eth_sendRawTransaction` requests) so will give an inaccurate response when the `"pending"` block is queried and the user has any pending private transactions.

### Desired Behavior

When a user sends the `eth_getTransactionCount` request to the RPC endpoint targeting the `"pending"` block the response should take into consideration any outstanding private transactions.

#### Security Concerns

However, there's a security concern, in that making this data openly available potentially leaks information about usage of the private mempool.

- For example, a third party could continuously call `eth_getTransactionCount` targeting the account in question, and see when the user has submitted new transaction(s) to the private mempool.
- If the account in question has known usage patterns this information could be used in an attempt to execute a blind back run or front run on the account.

As such, we propose that information about private mempool transaction counts is only included when the request is signed with the same `X-Flashbots-Signature` method we use for signing relay requests, and that the users wallet key is used to sign the request.

## Implementation Logic

We propose the following implementation logic:

```mermaid
graph TD
    START[Start] --> METHOD{"RPC method is\neth_getTransactionCount?"}
    METHOD -->|Yes| PENDING{"params[1] is pending?"}
[...]

PR #151, opened 2024-08-05
feat: Return local nonce when getTransactionCount request is signed

ryanschneider:
## 📝 Summary

Implements an "MVP" version of #150: When a `eth_getTransactionCount` RPC request is signed with a `X-Flashbots-Signature` header matching the target address we return the "next nonce" from Redis instead of proxying the request.

## ⛱ Motivation and Context

This feature is desirable for Wallets that want to use `eth_getTransctionCount` to manage user nonces while still preserving users' privacy.

Note that this is the _MVP_ implementation on this functionality, in that it uses the existing `RState.GetSenderMaxNonce(addr)` check.  As noted in #150 this current code does not respect "nonce gaps" so a follow-on PR will be needed later to correctly handle nonce gaps (most likely via adding a per-address Redis Sorted Set ("ZSet") as mentioned in the issue).  However, this implementation will work for most scenarios, and will allow us to get feedback from wallets before working on the more-detailed implementation.

## 📚 References

Tested as follows:

```
# Check an address nonce
> cast nonce 0x2485aaa7c5453e04658378358f5e028150dc7606 --rpc-url https://rpc-sepolia.flashbots.net --block pending
54

#Send a private tx from this address
cast send .... --rpc-url https://rpc-sepolia.flashbots.net

#Confirm cast nonce still returns the same value
> cast nonce 0x2485aaa7c5453e04658378358f5e028150dc7606 --rpc-url https://rpc-sepolia.flashbots.net --block pending
54

# Call the RPC w/ the X-Flashbots-Signature (code snippet)
responseBytes := sendSignedRPC(t, url, map[string]interface{}{
			"jsonrpc": "2.0",
			"id":      1,
			"method":  "eth_getTransactionCount",
			"params":  []interface{}{"0x2485aaa7c5453e04658378358f5e028150dc7606", "pending"},
		})
{"id":1,"result":"0x37","jsonrpc":"2.0"}

# Once the private tx is mined, confirm both cast nonce and the RPC return 0x37
> cast --to-hex $(cast nonce 0x2485aaa7c5453e04658378358f5e028150dc7606 --rpc-url https://rpc-sepolia.flashbots.net --block pending)
0x37

responseBytes := sendSignedRPC(t, url, map[string]interface{}{
			"jsonrpc": "2.0",
			"id":      1,
			"method":  "eth_getTransactionCount",
			"params":  []interface{}{"0x2485aaa7c5453e04658378358f5e028150dc7606", "pending"},
		})
{"id":1,"result":"0x37","jsonrpc":"2.0"}
```

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/adapters/flashbots/signature.go
+++ b/adapters/flashbots/signature.go
@@ -16,7 +16,7 @@ var (
 	ErrInvalidSignature = errors.New("invalid signature provided")
 )
 
-func ParseSignature(header string, body []byte) (string, error) {
+func ParseSignature(header string, body []byte) (signingAddress string, err error) {
 	if header == "" {
 		return "", ErrNoSignature
 	}
@@ -29,7 +29,7 @@ func ParseSignature(header string, body []byte) (string, error) {
 	return VerifySignature(body, splitSig[0], splitSig[1])
 }
 
-func VerifySignature(body []byte, signingAddressStr, signatureStr string) (string, error) {
+func VerifySignature(body []byte, signingAddressStr, signatureStr string) (signingAddress string, err error) {
 	signature, err := hexutil.Decode(signatureStr)
 	if err != nil || len(signature) == 0 {
 		return "", fmt.Errorf("%w: %w", ErrInvalidSignature, err)
--- a/adapters/flashbots/signature.go
+++ b/adapters/flashbots/signature.go
@@ -1,0 +1,67 @@
+// Package flashbots provides methods for parsing the X-Flashbots-Signature header.
+package flashbots
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+var (
+	ErrNoSignature      = errors.New("no signature provided")
+	ErrInvalidSignature = errors.New("invalid signature provided")
+)
+
+func ParseSignature(header string, body []byte) (string, error) {
+	if header == "" {
+		return "", ErrNoSignature
+	}
+
+	splitSig := strings.Split(header, ":")
+	if len(splitSig) != 2 {
+		return "", ErrInvalidSignature
+	}
+
+	return VerifySignature(body, splitSig[0], splitSig[1])
+}
+
+func VerifySignature(body []byte, signingAddressStr, signatureStr string) (string, error) {
+	signature, err := hexutil.Decode(signatureStr)
+	if err != nil || len(signature) == 0 {
+		return "", fmt.Errorf("%w: %w", ErrInvalidSignature, err)
+	}
+
+	if signature[len(signature)-1] >= 27 {
+		signature[len(signature)-1] -= 27
+	}
+
+	hashedBody := crypto.Keccak256Hash(body).Hex()
+	messageHash := accounts.TextHash([]byte(hashedBody))
+	signaturePublicKeyBytes, err := crypto.Ecrecover(messageHash, signature)
+	if err != nil {
+		return "", fmt.Errorf("%w: %w", ErrInvalidSignature, err)
+	}
+
+	publicKey, err := crypto.UnmarshalPubkey(signaturePublicKeyBytes)
+	if err != nil {
+		return "", fmt.Errorf("%w: %w", ErrInvalidSignature, err)
+	}
+	signaturePubkey := *publicKey
+	signaturePubKeyAddress := crypto.PubkeyToAddress(signaturePubkey).Hex()
+
+	// case-insensitive equality check
+	if !strings.EqualFold(signaturePubKeyAddress, signingAddressStr) {
+		return "", fmt.Errorf("%w: signing address mismatch", ErrInvalidSignature)
+	}
+
+	signatureNoRecoverID := signature[:len(signature)-1] // remove recovery id
+	if !crypto.VerifySignature(signaturePublicKeyBytes, messageHash, signatureNoRecoverID) {
+		return "", fmt.Errorf("%w: %w", ErrInvalidSignature, err)
+	}
+
+	return signaturePubKeyAddress, nil
+}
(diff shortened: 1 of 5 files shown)


server/request_handler.go
```diff
@@ -134,11 +134,11 @@ func (r *RpcRequestHandler) process() {
 	r.logger = r.logger.New("rpc_method", jsonReq.Method)
 
 	// Process single request
-	r.processRequest(client, jsonReq, origin, referer, isWhitehatBundleCollection, whitehatBundleId, urlParams, r.req.URL.String())
+	r.processRequest(client, jsonReq, origin, referer, isWhitehatBundleCollection, whitehatBundleId, urlParams, r.req.URL.String(), body)
```
metachris, line 137: at some point we need to start passing in a struct :)

adapters/flashbots/signature.go
```diff
@@ -0,0 +1,67 @@
+// Package flashbots provides methods for parsing the X-Flashbots-Signature header.
```
metachris, line 1: Tests for this would be great 🤠

server/request_processor.go
```diff
@@ -58,7 +60,7 @@ func NewRpcRequest(
 	rpcCache *application.RpcCache,
 ) *RpcRequest {
 	return &RpcRequest{
-		logger:                     logger,
+		logger:                     logger.With("method", jsonReq.Method),
```
metachris, line 63: nice

adapters/flashbots/signature.go
```diff
@@ -0,0 +1,67 @@
...
+
+	splitSig := strings.Split(header, ":")
+	if len(splitSig) != 2 {
+		return "", ErrInvalidSignature
+	}
+
+	return VerifySignature(body, splitSig[0], splitSig[1])
+}
+
+func VerifySignature(body []byte, signingAddressStr, signatureStr string) (string, error) {
```
metachris, line 32: explicit names for the return values would increase clarity 

```suggestion
func VerifySignature(body []byte, signingAddressStr, signatureStr string) (signaturePubkeyHex string, err error) {
```

server/request_handler.go
```diff
@@ -134,11 +134,11 @@ func (r *RpcRequestHandler) process() {
 	r.logger = r.logger.New("rpc_method", jsonReq.Method)
 
 	// Process single request
-	r.processRequest(client, jsonReq, origin, referer, isWhitehatBundleCollection, whitehatBundleId, urlParams, r.req.URL.String())
```
metachris, line 137: I wonder if a better approach would be not to pass the full body around for possible later signature check, but instead here do a signature check if the header is present, and only pass the result around.

Downside of that approach: doing the signature check also on requests that wouldn't need it. Upside: everything is a bit simpler.

server/request_handler.go
```diff
@@ -147,6 +147,11 @@ func (r *RpcRequestHandler) processRequest(client RPCProxyClient, jsonReq *types
 	}
 	// Handle single request
 	rpcReq := NewRpcRequest(r.logger, client, jsonReq, r.relaySigningKey, r.relayUrl, origin, referer, isWhitehatBundleCollection, whitehatBundleId, entry, urlParams, r.chainID, r.rpcCache)
+
+	if r.req.Header.Get("X-Flashbots-Signature") != "" {
```
metachris, line 151: i don't love this (passing signature and body around as part of the request), but i can see why we'd do it this way -- to only verify the signature on the methods that need it. so all in all it's also okay for me.

server/request_intercepts.go
```diff
@@ -150,3 +152,48 @@ func (r *RpcRequest) intercept_eth_call_to_FlashRPC_Contract() (requestFinished
 	r.logger.Info("Intercepted eth_call to FlashRPC contract")
 	return true
 }
+
+func (r *RpcRequest) intercept_signed_eth_getTransactionCount() (requestFinished bool) {
```
metachris, line 156: I would love if we could start doing unit / e2e tests for these handlers, but that seems def out of scope for now.

metachris:
LGTM! Left a few minor comments. 

Most notably, please add tests to the signature methods.

I don't love the name `adapters` when it's really just utils, but not having a strong opinion on it and it's fine as is.

server/request_handler.go
ryanschneider: Ya I like that too, I'll make that change and add tests for the signatures.

server/request_intercepts.go
```diff
@@ -150,3 +152,48 @@ func (r *RpcRequest) intercept_eth_call_to_FlashRPC_Contract() (requestFinished
...
+	signingAddress, err := flashbots.ParseSignature(r.flashbotsSignature, r.flashbotsSignatureBody)
+	if errors.Is(err, flashbots.ErrNoSignature) {
+		r.logger.Info("[eth_getTransactionCount] No signature found")
+		return false
+	}
+
```
ryanschneider, line 162: Oh wow just realized when refactoring this isn't handling the else case and should be returning an error result.  The refactor I'm doing to move the signature check earlier in the flow helps make that more obvious 😊

server/request_handler.go
ryanschneider: Agreed. :). Luckily the refactor you suggested, which is implemented in https://github.com/flashbots/rpc-endpoint/pull/151/commits/d90fb9b1c8de38358dbf4e5c748e52b234462300 removes the additional parameter.

adapters/flashbots/signature.go
ryanschneider: Add in https://github.com/flashbots/rpc-endpoint/pull/151/commits/d90fb9b1c8de38358dbf4e5c748e52b234462300, still need to write unit tests those are coming next.

server/request_handler.go
ryanschneider: Refactored to that approach in https://github.com/flashbots/rpc-endpoint/pull/151/commits/d90fb9b1c8de38358dbf4e5c748e52b234462300.

server/request_intercepts.go
ryanschneider: Fixed in https://github.com/flashbots/rpc-endpoint/pull/151/commits/d90fb9b1c8de38358dbf4e5c748e52b234462300.

adapters/flashbots/signature.go
ryanschneider: Added in https://github.com/flashbots/rpc-endpoint/pull/151/commits/e4f8b3119d455f0ea0e8bc75d3859a883c496e76!

PR #152, opened 2024-08-14
chore(signature): Double-check signature logic using third-party tools.

ryanschneider:
## 📝 Summary

Adds two more unit tests for `flashbots.ParseSignature` using two signatures generated via MetaMask and `cast`.

## ⛱ Motivation and Context

I realized that the existing `TestParseSignature` test uses the `crypto` package to generate the signatures, and since this is the same package we use internally to verify the signatures it was possible that we had a bug in both our uses for generation and validation of the package.  So as an extra sanity check I generated two additional signatures using MetaMask and `cast` to confirm that our signature verification code works across multiple signature implementations.

## 📚 References

I actually wasn't aware of this nice extra feature of etherscan where signatures can be published that also offers a nice UI for prompting your wallet to generate a signature: https://etherscan.io/verifiedSignatures#

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/adapters/flashbots/signature_test.go
+++ b/adapters/flashbots/signature_test.go
@@ -2,6 +2,7 @@ package flashbots_test
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/accounts"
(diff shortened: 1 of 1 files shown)


metachris (approved):
nice

PR #162, opened 2024-11-24
fix: rpc_endpoint_requests insert sql

ohko4711:
## 📝 Summary
fix: https://github.com/flashbots/rpc-endpoint/issues/161

## ⛱ Motivation and Context

rpc_endpoint_requests don't have filed `ip_hash`, save to db will failed
```
CREATE TABLE rpc_endpoint_requests(
    id uuid not null unique primary key default gen_random_uuid(),
    received_at timestamp with time zone not null,
    inserted_at timestamp with time zone not null default now(),
    request_duration_ms bigint not null,
    is_batch_request boolean,
    num_request_in_batch integer,
    http_method varchar(10) not null,
    http_url varchar(100) not null,
    http_query_param text,
    http_response_status integer,
    origin text,
    host text,
    error text
);
```

## 📚 References

---

* [ ] `make lint`
* [ ] `make test`
* [ ] `go mod tidy`

### DIFF
--- a/database/postgres_store.go
+++ b/database/postgres_store.go
@@ -32,7 +32,7 @@ func (d *postgresStore) Close() {
 
 func (d *postgresStore) SaveRequestEntry(entry RequestEntry) error {
 	query := `INSERT INTO rpc_endpoint_requests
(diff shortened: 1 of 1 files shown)


ohko4711:
@metachris  sir, can you take a look at this?  That way I won't have to maintain my own local version

PR #166, opened 2025-02-07
fix(nonce): Reset user nonce if tx fails.

ryanschneider:
## 📝 Summary

With the addition of #151 the rpc-endpoint can return the cached nonce if the `eth_getTransactionCount` request is signed.

However, this didn't take into account the fact that txs can fail, in which case we want to invalidate the cache.

Also, with the new `blockRange` URL param we want to shorten the nonce cache duration to match.

## ⛱ Motivation and Context

Say a user sends a tx w/ nonce 100, and it fails.  Then sends a signed `eth_getTransactionCount` request within the default nonce expiration window.  We would tell the user to use nonce 101 when really they should send a new tx w/ 100 since that tx never made it on-chain.

We relay on the fact that wallets send `eth_getTransactionReceipt` to check the tx status, and use piggy back off the existing logic already added in #13 by also clearing the cached max nonce if the tx fails.

## 📚 References

---

* [x] `make lint`
* [x] `make test` 
* [x] `go mod tidy`

### DIFF
--- a/server/request_intercepts.go
+++ b/server/request_intercepts.go
@@ -37,6 +37,28 @@ func (r *RpcRequest) check_post_getTransactionReceipt(jsonResp *types.JsonRpcRes
 		return false
 	}
 
+	resetMaxNonce := func(txFrom string, txHash string) {
+		// if the tx failed then we want to reset the redis max nonce
+		maxNonce, found, err := RState.GetSenderMaxNonce(txFrom)
+		if err != nil {
+			r.logger.Error("[post_getTransactionReceipt] GetSenderMaxNonce failed", "error", err)
+			return
+		}
+
+		// if the user doesn't have a cached nonce then we can exit early
+		if !found {
+			return
+		}
+
+		// we can elide error checking here since a txNonce of 0 will never match
+		txNonce, _, _ := RState.GetNonceOfTxHash(txHash)
+		if maxNonce == txNonce {
+			if err := RState.DelSenderMaxNonce(txFrom); err != nil {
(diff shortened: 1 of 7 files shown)


server/request_intercepts.go
```diff
@@ -37,6 +37,28 @@ func (r *RpcRequest) check_post_getTransactionReceipt(jsonResp *types.JsonRpcRes
...
+		// if the user doesn't have a cached nonce then we can exit early
+		if !found {
+			return
+		}
+
+		// we can elide error checking here since a txNonce of 0 will never match
+		txNonce, _, _ := RState.GetNonceOfTxHash(txHash)
+		if maxNonce == txNonce {
+			if err := RState.DelSenderMaxNonce(txFrom); err != nil {
+				r.logger.Error("[post_getTransactionReceipt] DelSenderMaxNonce failed", "error", err)
```
TymKh, line 57: It will work properly in like 99% of the cases if we have single pending tx and it failed. We are not trying to address all the complexities of multiple stacked txs right?
We probably at some point want to put all the nonce related info in single redis object per `from` address so that we could atomically act on it

TymKh (approved):
Thanks a lot for the fix, I think it will address current issue. I don't really like overall all the patchy things we are doing for nonce management, we should probably rework it at some point.

ryanschneider:
Thanks @TymKh I agree that the patch work approach isn't ideal, see my original proposed "Redis Sorted Set Implementation" at the bottom of #150 which I still think is the best way to handle this correctly.

PR #181, opened 2025-07-18
feat: add origin keyword support for refund configuration

odysseus0:
## Summary
- Added support for "origin" keyword in refund configuration URLs
- Allows wallets to use static URLs like `/endpoint?refund=origin:50&refund=0xWalletAddress:50`
- The "origin" keyword automatically resolves to the transaction sender address

## Problem
Currently, wallets need to manually update RPC URLs with user addresses for each request when splitting refunds. This creates unnecessary complexity for wallet integrations.

## Solution
Implemented a keyword system where "origin" can be used as a placeholder that gets replaced with the actual transaction sender address during request processing.

## Implementation Details
1. Added `IsKeyword` field to `RefundConfig` struct (non-serialized to maintain API compatibility)
2. Updated URL parameter parsing to recognize "origin" keyword (case-insensitive)
3. Added keyword replacement logic in request processor before sending to relay
4. Included proper error handling with helpful messages for unsupported keywords

## Testing
- Added unit tests for URL parameter parsing with origin keyword
- Added integration test verifying full transaction flow
- All existing tests pass

## Backwards Compatibility
- No breaking changes to existing API
- `IsKeyword` field is not serialized to JSON (`json:"-"`)
- Existing refund configurations continue to work as before

### DIFF
--- a/server/request_processor.go
+++ b/server/request_processor.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/flashbots/rpc-endpoint/adapters/flashbots"
 	"github.com/flashbots/rpc-endpoint/application"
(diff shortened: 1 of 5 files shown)


odysseus0:
Great suggestion @jeyldii\! 🎯 

I've implemented the change to use null address as a sentinel value instead of the `IsKeyword` field. This definitely simplifies the implementation by:
- Removing the extra struct field 
- Eliminating the need for the `isKeywords` array
- Using a well-understood Ethereum pattern

The refactoring is complete and all tests are passing. Thanks for helping me learn the team's coding preferences\!

server/url_params.go
```diff
@@ -131,18 +132,26 @@ func ExtractParametersFromUrl(reqUrl *url.URL, allBuilders []string) (params URL
 			if len(split) != 2 {
 				return params, ErrIncorrectRefundQuery
 			}
-			if !common.IsHexAddress(split[0]) {
-				return params, ErrIncorrectRefundAddressQuery
+
+			addressPart := strings.ToLower(split[0])
+			if addressPart == "origin" {
+				addresses[i] = common.Address{}
+			} else if !common.IsHexAddress(split[0]) {
```
TymKh, line 139: This if->elif is overcomplicated, let's keep logic on the first depth level:
It could be simplified to:
if !common.IsHexAddress(split[0]) && split[0]!="origin" { return params, ErrIncorrectRefundAddressQuery}
if split[0]=="origin" {
   split[0] == "0x00000..."
}
These 3 lines should suffice. Also it would be great to move "origin" and "0x0000..." to consts

server/request_processor.go
```diff
@@ -297,8 +298,37 @@ func (r *RpcRequest) sendTxToRelay() {
 	}
 
 	sendPrivateTxArgs := types.SendPrivateTxRequestWithPreferences{}
+
+	// Check if we need to replace the null address (origin placeholder)
+	needsKeywordReplacement := false
```
TymKh, line 303: Let's extract function:
replaceZeroRefund(refunds []types.RefundConfig, sender common.Address){}
that will replace 0x000 address with sender. Sender could be taken, by parsing r.txFrom.
Everything else then will be kept in place, thus it will be the least invasive change

server/url_params.go
odysseus0: Thanks @TymKh\! I've flattened the logic and added the constant in commit 90ee289.

I kept the distinction between wrong addresses and wrong keywords:
- `refund=0x123:50` → "Incorrect refund address"
- `refund=sender:50` → "Unsupported refund keyword (only 'origin' is supported)"

Your simpler approach would work too - it would return "Incorrect refund address" for both cases. 

What's the team preference here - simpler code or more specific error messages? Happy to go either way\!
EOF < /dev/null

server/request_processor.go
odysseus0: Thanks @TymKh! I've extracted the function as suggested in commit c97b438.

Initially I was being conservative about preserving the original URL parameters (in case they were needed for debugging/logging), but you're right that it's unnecessary here. The simplified implementation is much cleaner - just a simple function call that replaces null addresses in place.

All tests pass!

jeyldii (approved):

PR #182, opened 2025-08-20
feat: eth_sendPrivateTransaction

shanejonas:
## 📝 Summary

This adds support for `eth_sendPrivateTransaction` instead of letting it fall through to the provider.

## ⛱ Motivation and Context

Upstream provider is deprecating their support for this method, and adds an extra hop.

## 📚 References

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/server/request_sendprivatetx.go
+++ b/server/request_sendprivatetx.go
@@ -1,0 +1,37 @@
+package server
+
+import (
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/flashbots/rpc-endpoint/types"
+)
+
+func (r *RpcRequest) handle_sendPrivateTransaction() {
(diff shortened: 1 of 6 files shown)


server/request_sendprivatetx.go
```diff
@@ -0,0 +1,37 @@
...
+				r.urlParams.pref.Fast = true
+			} else {
+				r.writeRpcError("Preferences fast must be a boolean", types.JsonRpcParseError)
+				return
+			}
+		} else {
+			r.writeRpcError("Preferences must be an object", types.JsonRpcParseError)
+			return
+		}
+	}
```
shanejonas, line 34: still not happy with how nested this is

TymKh (approved):

